Issue #106, opened 2011-12-03
Add Multilanguage support


pokoli:
Mezzanine it's a cool django based cms but IMHO lacks of multilanguage support. 

Why not integrating a django-app for translating like: 

http://code.google.com/p/django-transmeta/

or

http://code.google.com/p/django-modeltranslation/

Lumpusz:
It's been almost a year now. Is there any plans on adding support for multilingual sites based on mezzanine? @stephenmcd Do you have plans to merge any of the solutions implemented by others, or planning something on your own? Please, give us some update, as multilingual support would be vital for many of us! Thanks.

stephenmcd:
Last I recall it was incomplete. 

Would you be interested in reviving it and working out what needs to be done and finishing it off?

Lumpusz:
Well... To be honest I've only started development with mezzanine recently. But sure, why not! Do you have some kind of checklist what works and what is missing? Are you planning to work on it yourself?

stephenmcd:
No on both accounts I'm afraid, which is why I mentioned the need to first work out what needs to be done :-)

I'd suggest reading through the linked PRs in this thread as well as the mailing list thread to start with. I vaguely recall it being complete apart from the admin interface which was showing all fields for all languages at once, and needed some form of JS/tabbed view for showing fields for only one language at a time. Perhaps others who have been involved can provide some more detailed guidance than that.

alan-hicks:
From my limited experience clients tend to fall into two categories, one where content is merely translated, and the other more language friendly one where content is broadly mirrored on language micro-sites.

The fork at https://github.com/alan-hicks/mezzanine does the latter with a working demo at http://mldemo.p-o.co.uk/ and source at https://github.com/alan-hicks/mldemo there is also a more realistic work in progress at http://raitbattery.p-o.co.uk/

Lumpusz:
@stephenmcd I've just found such a widget here https://github.com/stephenmcd/mezzanine/issues/106#issuecomment-30782121. Don't you think that can be used as the missing part?
[...]

PR #1019, opened 2014-04-27
Integrates django-modeltranslation as an optional app à la south

Kniyl:
I finally decided to take the time to integrate my wrapper shown in #106 into mezzanine. The code is far better that way.

One downside though:

```
======================================================================
ERROR: test_mulisite (mezzanine.core.tests.CoreTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/fields/related.py", line 303, in __get__
    rel_obj = getattr(instance, self.cache_name)
AttributeError: 'RichTextPage' object has no attribute '_page_ptr_cache'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mathias/PkgBuilding/mezzanine_devel/mezzanine/mezzanine/core/tests.py", line 238, in test_mulisite
    site1.delete()
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/base.py", line 694, in delete
    collector.collect([self])
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/deletion.py", line 197, in collect
    field.rel.on_delete(self, field, sub_objs, self.using)
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/deletion.py", line 17, in CASCADE
    source_attr=field.name, nullable=field.null)
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/deletion.py", line 197, in collect
    field.rel.on_delete(self, field, sub_objs, self.using)
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/deletion.py", line 17, in CASCADE
    source_attr=field.name, nullable=field.null)
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/deletion.py", line 181, in collect
    parent_objs = [getattr(obj, ptr.name) for obj in new_objs]
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/deletion.py", line 181, in <listcomp>
    parent_objs = [getattr(obj, ptr.name) for obj in new_objs]
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/fields/related.py", line 320, in __get__
    rel_obj = qs.get()
  File "/home/mathias/PkgBuilding/mezzanine_devel/lib/python3.4/site-packages/django/db/models/query.py", line 307, in get
    self.model._meta.object_name)
mezzanine.pages.models.DoesNotExist: Page matching query does not exist.
```

I tried to investigate why this is happening but I lack knowledge about the internals of mezzanine and django-modeltranslation. It can be that when models are registered for translation, their managers are not correctly patched but I’m very unsure about this.

Anyway, this issue has a workaround since it happens only when one tries to delete a ```Site``` with remaining ```Pages``` attached to it. If all pages are deleted beforehand, the ```Site``` delete smoothly.


fguimara:
Any on this front? Is it supposed to be any merge soon?

jerivas:
Looking very good! The only issue I see right now (at least on the Admin side) is that inlines are not tabbed (for example, in Galleries and Forms). I'm not sure if Model Translation provides classes for them.

Lumpusz:
Hi Mathias, thank you for taking care of this long awaited feature request! Looking forward for this work to me merged into the official mezzanine repository.

jerivas:
So, off the top of my head, this is the list of things that need to be done:

- [ ] Add some language switching mechanism to Inlines in the Admin.
- [ ] Add some language switching mechanism to localizable settings in the Settings Admin.
- [ ] Catch up with `master` to account for all changes in Mezzanine so far.
- [ ] Write tests for translation related stuff.

@stephenmcd, what do you think about these points? Anything else that needs to be done before merge? :)

jerivas:
Just for reference, Model Translation DOES include Admin classes for inline Admins. https://django-modeltranslation.readthedocs.org/en/latest/admin.html#admin-inlines

Kniyl:
It does but these are not tabbed. Tabbed ones are only regular ones : https://django-modeltranslation.readthedocs.org/en/latest/admin.html#tabbed-translation-fields-admin-classes

Maybe a feature request to the modeltranslation team will avoid us to duplicate what they could come up with some day.

jerivas:
Hmm, according to this [issue in model-translation](https://github.com/deschler/django-modeltranslation/issues/66) tabbed inlines are supported (both the stacked and tabular flavors). Perhaps Mezzanine's custom inlines are messing something up?

Lumpusz:
I am happy with non tabbed inlines, as long as I§m able to create multilingual sites.. If we use django-modeltranslation the models can be generated with non tabbed inlines. If later on django-modeltranslation gets support for tabbed inlines we can migrate easier because we don't need to touch the db schema. We may need to replace the admin classes but it is far less painful, than migrating a db schema.
@Kniyl but I agree that anything related to customizing modeltranslation should be done in that project.

stephenmcd:
@jerivas that checklist looks correct. As always, thanks your help here.

Lumpusz:
Regarding slugs: My take on this is that we only need slugs generated in the default language and use i18n_patterns in the url to differentiate between languages. Just the way it is done on the admin site, and in django. If we postpone the decision and go with a temporary one, which may change in the future, we also need to take care of backward compatibility, i.e.: providing redirects for old urls to new ones, to overcome potential broken external backlinks. Just keeping SEO in mind.

What are the pros and cons of implementing one solution over the other?

@jerivas @stephenmcd there might be another item to your list, and that is enhancing sitemap generation. https://support.google.com/webmasters/answer/189077?hl=en. (Again thinking of SEO)

Kniyl:
@jerivas You’re right. I just made a new django site with modeltranslation and both inlines are tabbed. Thus the issue seems to be with mezzanine or perhaps grappelli. I’ll try to figure out what are the requirements for modeltranslation that exist in plain django but not in mezzanine.

Kniyl:
@Lumpusz both translated slugs and slugs generated in the default language registered with i18n_patterns are valid approaches. The latter, however is clearly the easiest one since there is nothing more to do. If someone wants to include slugs inside i18n_patterns, it is also their responsibility to add hreflang support in their templates/base.html

If the choice is to provide translated slugs, then I agree that hreflang support should be considered. There are, however, two main drawbacks to this approach : settings.BLOG_SLUG (which might be solved by using a tupple or a dictionnary instead of a single string) and slug = / required to have the homepage within the page tree.

I have the same feeling, slugs in the default language + i18n_patterns if needed (which is already the case) seems safer.

Lumpusz:
@Kniyl Sorry I meant hreflang support in sitemaps, here is the correct link https://support.google.com/webmasters/answer/2620865?hl=en.

@stephenmcd, @jerivas  do you have anything against going with the slugs to be generated in the default language, and using i18n_patterns?

Kniyl:
I had a feeling that tabbed inlines disapeared because of grappelli and it seems to be it.

The way ```modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js``` tracks translation fields and adds the tabs to them is heavily dependent on the plain django templates. When translation fields are found (and especially for inlines) they track their container with ```$(el).closest('.form-row')``` which django uses as a class for each row in tabular inlines. They also track the ```table``` element to get rid of duplicates titles.

Unfortunately, grappelli_safe got rid of both. You can convince yourself by editing ```grappelli_safe/templates/admin/edit_inline/tabular.html``` and add ```form-row``` to the ```class``` attribute of the ```div``` at line 38. TabularInlines will now display translation tabs but they will get messy and disapear. And titles will still display for each languages.

So we have two options to get tabbed inlines. Either we change grappelli_safe to use a ```table``` instead of a ```div``` at line 11 and the ```form-row``` class at line 38 and whatever else is needed to get modeltranslation’s javascript to work, or we duplicate ```tabbed_translation_fields.js``` into mezzanine and we modify it for the hierarchy we are working with.

Imo, the second one seems more interesting and less painfull. I’ll try to come up with something in the near future. And may be I’ll get two birds with one stone and tackle the tabbed settings issue too.

jerivas:
@Lumpusz I would prefer translatable slugs for consistency. However, this is merely a personal opinion and in the end is Steve's decision to make.

@Kniyl That's good news. Let us know when you've got something and I will give it a test run.

stephenmcd:
If translatable slugs is what someone wants, then really wants the point of all this? Why not just construct a page tree with the primary nodes representing each language, the secondary nodes representing the primary pages per language, and so forth.

I honestly take this thinking further and question the need for translatable models at all - in my mind it actually adds a huge constraint to how a multi-lingual site works, in that *every* page needs to be translated in every language. Expanding on that - If the pages for each language don't have a relationship with each other, then you're no longer restricted by the above - one site may have extra pages that are only applicable to a particular language. Then the actual implementation of differentiation between languages on a site can occur via either a multi-site setup, eg: en.example.com, jp.example.com, etc, or via languages as root node in the page tree, eg example.com/en/, example.com/jp, etc.

Is the end result of integrating modeltranslation nothing but a nicer admin interface that allows you to edit multiple languages for a single page at once? Because given the restrictions described above and more importantly, the maintenance overhead and future complexity that this introduces to the project (which naturally isn't a primary concern for most of the people involved here), I really question how worthwhile it really is.

PS: This is why every time this and other similarly large features come up, I try and push for it to come together as a third-party package - that removes the maintenance overhead and complexity around future development as a concern, as well as the restrictions described above which would dictate "here's how to do multi-lingual" sites, should modeltranslation actually be integrated directly into Mezzanine.

jerivas:
I've used both approaches described by Steve for clients 
(language-specific subdomains and top-level nodes in the page tree for 
each language). I describe some of the drawbacks below (if you want all 
page trees in all languages to be identical):

  * Following any of the approaches described above puts the burden of
    keeping everything in "sync" between languages on the end user. If
    you add some content to a field in one language, you must remember
    to go to the "equivalent" page in the other languages and do the
    same. In fact, all of these actions must be manually repeated for
    every language you have: page creations, page removals, changes in
    sort order, changes in page menu options, etc. Model translation
    will do away with those.
  * Using the tree approach makes it complicated to leverage Django's
    language switching utilities, as you're not really using i18n
    "proper" and have to come up with a way to associate certain slugs
    with certain language.
  * Both of these approaches fail to provide easy "language switching"
    for visitors. If you land in a specific page of the site and it is
    not in your desired language, you choose to switch languages, but
    since there's no one-to-one correspondence between pages, you go to
    the home page of your desired language. Good luck hunting for the
    page you were looking for in the beginning.

Now, I empathize with Steve because I see how big of a maintenance 
burden this will be. Making this a third-party package sounds good to 
me, but some hooks will need to be setup in Mezzanine's core; like being 
able to choose from which model/admin classes to inherit for Mezzanine's 
Displayable, for example.

Lumpusz:
@jerivas I agree that language specific subdomains are not really an option for production sites, as they introduce way too much maintenance time/cost, and they are very prone to human errors. If we used the tree approach, we could use 99% of django's language switching mechanism to get back to the same translated page that the switching was originated from. According to the docs here https://docs.djangoproject.com/en/1.6/topics/i18n/translation/#the-set-language-redirect-view we can set the redirect url in the language switching form to be the translated version of the current url with slicing:
```html
<input name="next" type="hidden" value="{{ request.get_full_path|slice:'3:' }}" />
```
I'm also in favor of implementing this as an app, provided, that mezzanine implements the necessary hooks to use customized admin classes.

Kniyl:
Correct me if I’m wrong but isn’t modeltranslation an external app and this pull request the necessary hooks to have it integrated within mezzanine?

The whole point of this and #106 is to have an easy-to-use and already integrated solution to manage multi-lingual sites for people that want to. Reading all the comments asking for this kind of feature, I have the feeling that mezzanine’s newcomers don’t even think of the page tree as an alternative to i18n_patterns and, sooner or later, will look after this feature elsewhere. But, as far as this PR is concerned, we’re not forcing anyone to share django-modeltranslation’s view on how multi-lingual sites should be managed. It’s an optional feature that needs to be explicitly enabled by (1) installing django-modeltranslation and (2) activating it with ```settings.USE_MODELTRANSLATION = True```.

And, believe me, I actively try to keep the number of added/modified files to the minimum to reduce the maintenance burden. It is just that:
(1) There are some classes that need to be modified in order to greatly simplify future extensions (stephenmcd/cartridge#187 is really simple from this point of view).
(2) There were some requests that I interpreted as a requirement for the merge (language selector in the templates, translatable settings, tabbed admin) and it keeps adding to the number of modifications involved.
(3) I have a tendency to list every possible fix that comes to the top of my head when I identify an issue and it makes things feel like the solution will be a terrible blob. Whereas it is just my way of asking if I’m right considering one way simpler than the others.

That said, solutions exist to unattractively glue mezzanine and modeltranslation together as well the multi-site and page tree you mentionned. So there is a need of a clear statement of wether we close this PR, #106 and we document that multi-lingual support is not built in mezzanine but there are several solutions to accomplish this, or we continue on this path and make the integration as smooth as possible for those who consider they need this feature.

Lumpusz:
On the one hand mezzanine would become very attractive to people who want to build mutilingual cms sites, if this feature was integrated into mezzanine itself. On the other hand I understand the hesitation on boarding a new feature.
Considering all that stuff have been going on on the mail lists, forums, multilingual support is a long awaited feature (https://github.com/stephenmcd/mezzanine/issues/106 is opened >3 years). Merging this code into the master repo would mean finding the holy grail for many of us. I was just saying that a 3rd party app which mezzanine was tweaked to, would be better  than any of the solutions out there, if, and only if, the merge can't happen.
All that said I think if Admin classes and files are fairly independent from current codebase, the risk of merging the code is minimal, provided that appropriate test case coverage is also present. As @Kniyl said one can opt to use it or not.

Kniyl:
On a side note, I just merged stephenmcd/mezzanine:master into Kniyl/mezzanine:master and resolved the conflicts on my local repo. I’m 400 commits ahead and I’m wondering if they will all show up in here if I push the modifications on github.
Was it the way to go?

jerivas:
Yep, push it to to `Kniyl:master` and it'll show up here.

Kniyl:
Nice, I was affraid that all the 400 commits since my fork would show up too.

stephenmcd:
@jerivas Thanks for the feedback on my points, that all makes sense. I didn't want to sound like I was derailing the whole idea either, which I kind of felt like after my last comment, but I just wanted to play devil's advocate, particularly from a maintenance perspective.

jerivas:
No problem @stephenmcd, I'm glad we can consider all points of view and make a sound, well-established decision.

So, I went to the Travis build details and it looks like they are all passing (with the exception of one, but it looks like a network error), so I crossed off that item from the list. We still have three pending items. Is it safe to assume after those are completed a merge could happen? :)

stephenmcd:
Don't hold me to it, but sure, let's give it a try.

Kniyl:
I just pushed a version with tabs for tabular inline translation fields. It works and all but... it's kind of ugly when there are more than two or three languages (especially on a form page). I need some help to figure out what would be the best way to handle them : keep stacking them like they do but try to improve the layout or expand the cells to contain them all.

And... Did something change in the travis build config? The push fails with pep8 warnings on files that passed in the previous push and were left unchanged.

stephenmcd:
Awesome @Kniyl - I think we upgraded the pep8/pyflakes versions a while back, so yes there may be new regressions.

Are you able to post a screenshot or something of the tabbed interface?

Kniyl:
Sure! Here you go.

2015-02-09-212545_1366x768_scrot

Lumpusz:
@Kniyl This one seems to be a very good solution to organizing lots of tabs in small places: https://github.com/jasonday/plusTabs

stephenmcd:
You're right that's a bit of a mess.

Would it be better to somehow leverage the language selector for the admin interface? So choosing to use the admin itself in a particular language would then define the language you're also editing content for.

Kniyl:
@Lumpusz plusTabs doesn’t seems to work out of the box. I added plusTabs.min.js before tabbed_translation_fields.js in ```mezzanine.core.admin.BaseTranslationModelAdmin``` and changed each ```tabsContainer.tabs()``` of tabbed_translation_fields.js into
```
tabsContainer.plusTabs({
    className: "plus-tabs",
    seeMore: true,
    seeMoreText: "",
    showCount: false,
    expandIcon: "&#9660;",
    dropWidth: "66%",
    sizeTweak: 0
});
```
The result is... somehow less appealing than before.
2015-02-09-233415_1366x768_scrot

@stephenmcd Its worth investigating. I wouldn't choose the admin language as the default editing language since modeltranslation provides settings to exclude languages from translation. But it provides a global switch near the title of the admin form. It might be easier to use it.

Kniyl:
It is even simpler to just hide the tabs so that only the global switch is managing the language that shows up. And it also works when USE_L10N is set to False and the admin has a fixed language.

stephenmcd:
> It is even simpler to just hide the tabs so that only the global switch is managing the language that shows up. And it also works when USE_L10N is set to False and the admin has a fixed language.

Cool! That's kinda was I was thinking of.

Kniyl:
It should be it. Settings marked as translatable are now grouped and managed with a language switch as the other translation fields. It might need some visual improvement such as hinting which fields are duplicated in other languages, though. But as long as the ToDoList is concerned, there should only be the first item left.
And as far as I can tell re-reading the discussion, it seems that letting the user manage localized urls if they want to is an acceptable solution for anyone (with i18n_patterns being the obvious choice that works out of the box).

So... Nothing more needed?

jerivas:
Tried it out and it's working great. Only problem I see is that the language selector appears over the admin action buttons. Perhaps the action buttons need to be offset down or to the left to make room for the floating drop down.

stephenmcd:
Just tried it and noticed the same thing - but what I was getting at earlier was, do we really need separate language selectors in the admin for both the UI itself and content, and can we somehow have the translation fields leverage the UI admin selector?

jerivas:
Only downside I see to coupling both selectors (as Matthias noticed) is that model translation and the UI languages might not be the same, and some users might not enable L10N, which disables the UI lang selector.

Kniyl:
Thing is, I don't really know where to put this selector. I had it in the header next to the L10N one but it was confusing.
But I agree that its position should be improved. And, imo, it should have a fixed one so the users can access it when they are editing translation fields down below.

Kniyl:
I moved the selector down there. Is it better?
2015-02-10-215208_1366x768_scrot

stephenmcd:
Honestly I think it's a really confusing user experience to have two different language selectors, and more often than not the list of ui languages would match the list of content languages.

Kniyl:
I agree. And the fact that I don't really know where to put the selector to be both visible and clearly identifiable tends to confirm that.
However, there are currently two issues of using only one selector:
1. users who set ```USE_L10N``` to ```False``` won't be able to use a selector at all;
2. their behavior is completely different. The L10N one redirects to django's ```set_language``` on a click, discarding changes when the page is shown again; and the modeltranslation's one just hide/show some divs on the same page on a click.

1 can be easily solved by using ```{% if settings.USE_L10N or settings.USE_MODELTRANSLATION %}``` in the template instead of just ```{% if settings.USE_L10N %}```; but for 2... I still didn't have the epiphany to manage that.

Kniyl:
Ok… tried to get back on using tabs instead of the global switch to reduce confusion. I combined all tabs from TabularInline into a single one, though, so it is not ugly anymore. Does it seems good enought to be pushed or should I still try something else?

 * TabularInline: 
2015-02-11-224110_1366x768_scrot
 * Settings:
2015-02-11-223928_1366x768_scrot

Lumpusz:
@Kniyl this is the best so far. Do you think that instead of tabs, using radios, only for inlines, would look more appealing? http://jqueryui.com/button/#radio

Kniyl:
I rely on the activate (or select for jQuery-UI 1.8) event to trigger all hidden tabs at once. Unfortunately, buttons doesn't seems to provide such hooks.

stephenmcd:
I think the tabbed approach is still really poor from a user perspective. 

Visually the tabs are just hanging there and don't seem to flow from the field itself in a natural way, but please don't give that too much thought - the real problem here is that it's most likely that none of us are particularly great UI designers, so we're really just going around in circles here with all these attempts, in futility to some extent. 

More importantly, I think it's quite poor from a user interaction perspective. When I land on a form with 27 translatable fields, I now need to click 27 tabs just to be able to edit the content for one alternative language. If you think about how an actual translator will do this, they'll most likely be translating to a single language, one piece of content at a time - ideally there should only be one selection needed.

Both these issues are solved by the idea I proposed earlier, around leveraging the UI language selector. A translator would login to the system, and can actually use the language selector on the login screen to the same effect. Two issues were raised with this:

1) `USE_L10N` may not be configured. 

This is actually solved trivially. We pre-set a ton of different settings, based on the values of other settings, inside `mezzanine.utils.conf.set_dynamic_settings`. If `USE_MODELTRANSLATION` is set, we can set `USE_L10N` to True.

2) List of UI languages mightn't match translation languages.

I think more often than not this won't be the case. A project might have two or three translation languages, and presumably there'd be no real reason to not have those available as UI languages as well. And if there was an edge-case project where more UI languages were required than what can be translated, the translatable language could always fall back to a default in the case of a UI language being selected that translations aren't required for. Not sure how well I've described that, it feels a bit confusing, so I can certainly clarify further there if needed.

As for the actual implementation - surely it's not too much of a stretch to hide the translation selection widgets (tabs, dropdowns, whatever they currently are), and trigger default selection based on grabbing the current UI language?

Kniyl:
I’ve been thinking about this lately and I pretty much reached the same conclusion: L10N is not an issue and, since the translation languages should be a subset of the site languages, the differences between the two should not be difficult to handle.

But, the main issue that this approach gives is that the UI language selector triggers a redirect and can potentially lose data that weren't saved beforehand, unlike the selector from modeltranslation which hide/reveals fields on the same page.

I’ll push this solution, though. So we can check how much of a big deal it is compared to its advantages.

stephenmcd:
That data loss is a hugely important point - we could add some confirmation javascript to the language selector that warns the user and allows them to cancel changing the language, much like gmail does (or did) when you close the browser tab without sending an email that's been written.

jerivas:
Some alternatives off the top of my head, which would require creating a Mezzanine specific implementation of `set_language` view:

- Trigger save() before doing the redirect to avoid data loss. This will hit the DB every time the user changes language and might be slow.
- Persist the form data in the request (as POST data) or in the user's sessionStorage. Notice that `set_language` already uses POST data so we might not want to mess with it.

As a side note: @stephenmcd remember the global model translation switch is available in the current implementation, so a translator won't have to switch each field individually.

Kniyl:
I pushed a version that uses the active language to determine which fields to show/hide so you guys can try it and see how it feels. Nothing is done yet to handle the possible data loss.

But (as always, there is a but) a new problem shown up in the name of required fields. Modeltranslation does apply required attributes on the default language for fields that are defined with blank=False. Which means, the default language should be populated before any other otherwise validation errors appears while attempting to save a page. I’ll check later in the documentation of modeltranslation if it is possible to change this behavior but it breaks the natural flow of create, edit, save for whatever language is suitable to the user. However, one may argue that translators shouldn't have to create a page and must only edit those which already are saved with the default language.

Anyway, it adds to the "save" issue and we should consider wether we fix it and keep the nice looking interface or we get back to the terrible looking tabs which work out of the box.
Since modeltranslation won't be a default feature, my personal taste would be to use the tabs (and I think I can manage to get all the tabs synchronize without the need of a global switch, if needed), open an issue saying that the admin looks bad with modeltranslation enabled and hopefully someone with greater design skill will provide a solution.

I think I will cool my brain on that for the next few days, maybe it will help to come up with something that works best.

stephenmcd:
Thanks for all the effort @Kniyl - really awesome.

Kniyl:
The more I try to improve the "one language selector to rule them all" solution, the more I feel that tabs are the way to go. The data loss on language change and the required fields issue are too much of a hassle to handle properly. The approach provides a neat UI, though, and seems intuitive at first. I wish I had found a way to make it smooth for the user, because in its state it can be painful to use.

An other detail that tabs are improving is the ability to easily spot translatable fields. Having a bunch of language names on top of a field will make you know that you can provide translations for them even if you are not aware that the site is supposed to be multi-lingual.

As such, I pushed a version with a new design for tabs, hopping that they will feel better. This way, anyone can fork my repo and try to improve either the tabs version (last commit) or the language selector version (previous commit). This new version keeps in sync all tabs on a page so that when you change a language for a field, all other translatable fields are changed to the same language.

Example by image:
 * before click: 
2015-02-15-002958_1366x768_scrot
 * after click: 
2015-02-15-003002_1366x768_scrot

Lumpusz:
I think this approach combines the bests of all alternatives, of all discussed solutions (switching  language on a page with only one click, indicate translatable fields), and eliminates all the worst (confusion with multiple language selectors, as well as combining two similar, but not identical features under one selector.) It also looks better than the previos solution with tabs. Thanks @Kniyl!

Lumpusz:
@stephenmcd @Kniyl What is still missing for modeltranslation integration to be merged into master?

stephenmcd:
Free time.

stephenmcd:
Just reviewed this again and I think the UI works well.

One question - how does caching come into play? Currently with a production deploy, we enable caching of all pages using URLs as keys - it seems like the default setup with modeltranslation at the moment doesn't prefix URLs with language, eg foo.com/en/some-page, so would we end up with cached content for one language showing when another language has been selected?

We'll also need to squash the 30+ commits into one once this is ready, as each commit message gets put into the changelog.

jerivas:
I'll set up a site based on this branch on my VPS to test the caching. In the meantime, how does caching work with logged in users in both the public site and the admin? Because the URLs are the same, but depending on your credentials you'll see different things.

stephenmcd:
Cool Ed. Caching only comes into play for unauthenticated users, so no problem there.

jerivas:
Okay, it's up and running (http://dev.jerivas.com). As far as I can tell, pages are being correctly rendered for different languages. I've add this string to all templates, so you can see the last time Django parsed the templates: Generated at:  {% now "jS F Y H:i:s" %}.

stephenmcd:
Thanks Ed. I was surprised it worked, but looking at how the cache keys are generated, they actually take the current language into account, which totally explains it:

https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/utils/cache.py#L91

jerivas:
Huh, that's nice!

jerivas:
So nice to see this becoming part of Mezzanine :)

While getting the live installation ready, I noticed `createdb` will fail if you have `settings.USE_MODELTRANSLATION = True`. The workaround is to activate Model Translation only after the initial database installation, which then requires running migrations for the translation fields. Of course, I'm no Model Translation expert, so I wanted to ask @Kniyl what is your recommended approach to:

- Start a project with Model Translation serving several languages from the get-go
- Add Model Translation to an existing project

stephenmcd:
Thanks everyone who contributed to this, it's merged now - I've announced it here:

https://groups.google.com/forum/#!topic/mezzanine-users/imE1GEwlvLI

Kniyl:
@jerivas To start a Mezzanine project with modeltranslation integration from the start, one has just to set ```settings.USE_MODELTRANSLATION = True``` and provide at least two languages in ```settings.LANGUAGES``` and it’s done. However, as you said you had an error trying that, I checked again. Everything went fine with Django 1.6, but there is an issue with Django 1.7. Something to do with the manage command ```update_translation_fields``` being run too late, I guess. I’ll fix that as soon as possible.

As for adding modeltranslation afterwhile, one has to manually run ```manage.py update_translation_fields``` in addition to the two prerequisites mentionned above. And that’s it.

@stephenmcd How do I provide only one big commit ? rebase ?

twil:
Hi Stephen,

Thanks for the merge! I'm playing with mezzanine and modeltranslation the whole week now and it plays well. The only thing I can add this far is documentation needs mentioning such a usefull setting as `MODELTRANSLATION_FALLBACK_LANGUAGES`. Without it you could see blank posts.

Initially I had set

```python
LANGUAGES = (
    ('ru', _('Russian')),
    ('en', _('English')),
)
```

And the way modeltranslation works Russian became the default (fallback) language. Posts created in Russian were shown when viewing in both Russian and English but posts created in English shown blank lines when viewing site in Russian. After addition of 

```python
MODELTRANSLATION_FALLBACK_LANGUAGES = ('ru', 'en')
```

things settled and no blank lines are shown.

Kniyl:
@jerivas @stephenmcd The issue encountered when ```manage.py createdb``` is run under Django 1.7 has been fixed in Kniyl/mezzanine@6e03f1719eb01c7767bcb4a3b0ac098ec74fe451, should I make a new pull request for that ?

The issue was that, instead of creating tables from scratch, django 1.7 now creates tables out of migrations. Thus modeltranslation can't hook up and write its own column at creation time. ```manage.py sync_translation_fields``` has to be run before ```manage.py update_translation_fields``` and everything works fine.

stephenmcd:
Yes please do - thanks a lot!

tabatabaei:
when relase mezzanine with multilingual support ???
they are want to write multilingual extention for mezzanine  and  make multilingual page content, if empty  other language content translate from defult language content by google translator?

Kniyl:
I can't answer the first question since it depends on @stephenmcd schedule.

For the second one… can you rephrase it? I don't quite understand.

stephenmcd:
Next release discussed here: https://groups.google.com/d/msg/mezzanine-users/_QWfFVB3RVc/AgcYAjZ7fdkJ

PR #1101, opened 2014-08-20
Docs: Spanish translation.

jerivas:
This is still work in progress. I'm opening the issue just for comparison and review.


gual:
What's the current status with this, @jerivas ? I'll be glad to help.

PR #1190, opened 2015-01-03
nginx conf: use Mozilla recommended ciphers

lorin:
For the included nginx.conf, use the ciphersuite recommended by the [Operations Security team at the Mozilla project](https://wiki.mozilla.org/Security/Server_Side_TLS).

The ones included here are the ones that Mozilla labels "Intermediate compatibility". They also document a ciphersuite with a higher level of security on that page labelled "Modern compatibility", though that one excludes several older browsers.


stephenmcd:
Awesome, thanks Lorin.

PR #1191, opened 2015-01-04
Updated the js in footer_scripts.html to handle universal analytics

avery-laird:
I realized the js is out of date with google's new universal analytics script, updated it to match with the new standard.

### DIFF
--- a/mezzanine/core/templates/includes/footer_scripts.html
+++ b/mezzanine/core/templates/includes/footer_scripts.html
@@ -4,14 +4,12 @@
 
 <script>
 {% if settings.GOOGLE_ANALYTICS_ID and not request.user.is_staff %}
-var _gaq = _gaq || [['_trackPageview']];
-_gaq.unshift(['_setAccount', '{{ settings.GOOGLE_ANALYTICS_ID }}']);
-(function(d, t) {
-    var g = d.createElement(t),
-        s = d.getElementsByTagName(t)[0];
-    g.async = true;
-    g.src = '//www.google-analytics.com/ga.js';
-    s.parentNode.insertBefore(g, s);
-})(document, 'script');
+(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks Avery. 

You'll see in Cartridge that the old code (such as the `_gaq` variable) is leveraged in order to track the details of a completed purchase: https://github.com/stephenmcd/cartridge/blob/master/cartridge/shop/templates/shop/complete.html

Any idea what needs to happen there to use the newer code? We'll need to get that working before this can be merged.

Thanks again for your help.

AlexHill:
There is a bit of an issue here in that you have to have your Analytics property transferred to universal analytics before you upgrade your tracking code.

From the [Universal Analytics Upgrade Centre](https://developers.google.com/analytics/devguides/collection/upgrade/guide):

> Data sent from one of the new tracking libraries and protocols will not be processed by a property that hasn't been transfered to Universal Analytics.

Google is transferring properties automatically now, but I don't think it's compete yet so some people will still be on classic.

Alex

stephenmcd:
Thanks Alex.

Any idea if the old code will still work for accounts that have been transferred?

If so, I guess we can just keep the old code for now.

AlexHill:
The transition is [currently in the third of four phases](https://developers.google.com/analytics/devguides/collection/upgrade/index#phases). All new properties are Universal and the remaining legacy properties are being transferred. Once that's done, the old code will be considered deprecated and we'll be able to safely upgrade.

avery-laird:
Hi guys,

My bad, obviously I should have done a bit more homework! Glad you caught this before it became an issue.

AlexHill:
Don't sweat it Avery, this is work that would have needed to be done at some point – now this PR can just be put on ice until it's needed. Good stuff.

stephenmcd:
Yep no problem at all - definitely needed! Thanks Avery.

stephenmcd:
We now have the updated tracking code for Cartridge in https://github.com/stephenmcd/cartridge/issues/224 - anyone know if it makes sense to upgrade this now? Another way of putting it, is it impossible for some people to use the new code?

AlexHill:
> Universal Analytics is now the operating standard and all properties have been upgraded to Universal Analytics.

https://support.google.com/analytics/answer/4457764?hl=en

I've updated this patch at https://github.com/stephenmcd/mezzanine/tree/avery-laird-updated-google-UA to include mobile, and a `footer_js` block.

AlexHill:
Corresponding change in Cartridge: https://github.com/stephenmcd/cartridge/tree/updated-google-GA

AlexHill:
Closing this in favour of #1487. Thanks Avery.

PR #1192, opened 2015-01-08
If keywords do not save allow admin to force save

joshcartme:
If keywords fail to submit for some reason clicking any of the save buttons on an admin page will appear to do nothing.  This commit allows the admin user to choose whether or not to continue saving.  It also makes debugging easier as it makes apparent that there is a problem and what the problem is with.

I just ran into this myself when a client asked to change the admin url.  I changed it but initially forgot to add the new url to the SLL_SSL_FORCE_URL_PREFIXES setting.  In this case the site does use SSL so this broke the admin keywords submit because admin pages were insecure and posting to that secure URL.

You can simulate a broken keyword submit by just deleting everything that is inside the submitKeywords function.

I think that the keyword submit being broken for any reason shouldn't stop the rest of an admin form from saving.  This strikes a nice balance by allowing the user to continue saving, but still alerting them that there is a problem.

### DIFF
--- a/mezzanine/core/static/mezzanine/js/admin/keywords_field.js
+++ b/mezzanine/core/static/mezzanine/js/admin/keywords_field.js
@@ -56,6 +56,12 @@ jQuery(function($) {
             });
         };
         submitKeywords();
+        if (keywordsSaved == false){
+            var force_save = confirm("The keywords did not submit successfully, continue saving?");
+            if (force_save == true) {
+                return true;
+            }
+        }
         return false;
     });
 


stephenmcd:
This is really pragmatic but I don't think it's a good idea to introduce it.

It will only serve to hide underlying errors that need resolving, and possibly annoy users who end up having to hit the confirm dialogue. You then end up with a situation where the users end up living with the problem because it works well enough, and no one technical ends up working out what's really gone wrong.

If anything, I think we should put an alert dialogue here that says "There was an error storing keywords, your changes cannot be saved", and then not saving.

stephenmcd:
BTW thanks for bringing this up anyway - I think it does need addressing. Having the form silently fail is definitely worse than what you've proposed.

joshcartme:
Hey Steve, totally makes sense and I can see why you wouldn't want to take it as far as I did.  I'll update it to be an alert.

Thanks

joshcartme:
After playing around with this some more, my commit (or changing it to an alert) is actually broken.  The click function will always return false the first time which means the alert/confirm/whatever I put there will always be called.

It is only on the second call from button.click inside the asynchronous code that the function will return true and not hit the alert/confirm/whatever.  I had accidentally been covering this up with a breakpoint I was using to test but the end result is that the alert/confirm/whatever will show up every time if the async code doesn't run very quickly.

Do you have any thoughts on where to take this?  There is a fail handler for $.post, but I'm not sure if it would cover every case that we consider failure.  The problem that prompted me to look at this was that /admin_keywords_submit/ was returning a 307 since it secure and was being posted to from an insecure page.

stephenmcd:
Not sure, but the notion of having the global-ish `keywordsSaved` and calling `click` really smells.

What if instead of having `keywordsSaved`, once the form needs to be submitted we just get the form objects and call `submit)` on it.

joshcartme:
We may be able to do something like that, but just submitting the form won't be enough because then I think it will lose which submit button they pressed; Save, Save and Continue, etc... (similar to what happened at first when I recently worked on the dynamic inlines a bit).

stephenmcd:
Yeah you're right, nice pickup! Need to think about this more, it'd be great to clean it up, regardless of the error handling.

joshcartme:
I'll think about it some more as well.  Would you like me to file an issue and reference this to keep track of it?

stephenmcd:
That'd be good - I'll close this PR for now.

stephenmcd:
Thanks btw.

joshcartme:
No problem!

Issue #1159, opened 2014-11-23
ValueError when surfing to /comments/


Xeniac-at:
it's a minor inconvenience. 
When you try to open _http://your.mezzanine.page/comments/_ in your browser you will get a Error 500.
with _DEBUG = True_ it will be:

>  **ValueError** at /comment/
>  need more than 1 value to unpack
>  **Exception Location:**  /virtenv/lib/python2.7/site-packages/mezzanine/utils/models.py in get_model, line 35

Im pretty sure _/comment/_ is not meant to be accessed directly, but some bots try to do that.
I have error reporting via eMail enabled and recently i get a lot of this errors in my inbox.
I tried to reproduce it, and also Mezzanine Page is affected. see: http://mezzanine.jupo.org/comment/

IMHO accessing /comments/ via get should return an "**HTTP 405**    Method Not Allowed" Error.

btw: thank you for your great work on mezzanine.

mikeredbike:
I was getting the same thing when the crawler was looking for /rating/.  I suggest changing get_model on line 33 of mezzanine/utils/models.py to:
def get_model(app_label, model_name=None):          
    if model_name is None:                          
        if len(app_label.split('.')) != 2:          
            raise LookupError                       
        app_label, model_name = app_label.split('.')
    model = django_get_model(app_label, model_name) 
    if not model:                                   
        raise LookupError                           
    return model

PR #1194, opened 2015-01-12
Fix 500 error when 'comment' view function receives incomplete POST data

abbottc:
This is related to the problem expressed in issue #1159.

The view which responds to the "/comment/" URL doesn't validate the "content_type" parameter of the POST data it receives before processing it.  As a result, if POST is either missing the "content_type" parameter, or that parameter is malformed, then the view will produce an error.

The fix implemented for issue #1159 avoids the 500 error for GET requests by enforcing that the method only responds to POST.  However, the underlying code problem still exists.  A malicious bot that receives a 405 Method Not Allowed response to a GET request (as it now will for this URL), will typically read the response headers, and if POST is allowed, then attempt a POST request to continue probing the web server for spam opportunities or other weaknesses.

Under the current code, these POST requests will continue to produce 500 server errors.  The web server should respond with a 400 Bad Request response instead.  That's what this patch does.  A related test for the situation is included too.

### DIFF
--- a/mezzanine/generic/tests.py
+++ b/mezzanine/generic/tests.py
@@ -12,6 +12,7 @@ from mezzanine.core.models import CONTENT_STATUS_PUBLISHED
 from mezzanine.generic.forms import RatingForm
 from mezzanine.generic.models import (AssignedKeyword, Keyword,
                                       ThreadedComment, Rating)
+from mezzanine.generic.views import comment
 from mezzanine.pages.models import RichTextPage
 from mezzanine.utils.tests import TestCase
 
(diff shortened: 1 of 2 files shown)


stephenmcd:
What a great bug report and fix - perfect.

Thanks a lot!

abbottc:
Thanks for the quick merge, and thanks for all your work on this project.

edeesis:
I know this is a 3 year old PR, but I'm curious if either of you two had any insight into why this might be failing (note I am running an old version of Mezzanine that does not have this fix).

I'm getting this error inconsistently despite the request seemingly to have properly formed post data.

Here is the POST data we're getting (I get this every time this error happens):

`POST:<QueryDict: {u'comment': [u"****"], u'name': [u'******'], u'url': [u''], u'referrer': [u'******''], u'object_pk': [u'5859'], u'security_hash': [u'******''], u'content_type': [u'blog.blogpost'], u'timestamp': [u'1537106569'], u'honeypot': [u''], u'csrfmiddlewaretoken': [u'******''], u'email': [u'******'']}>`

I haven't been able to figure out the root cause of this because it's not happening on my accounts and as far as I can tell from reading the code, and debugging through myself, everything looks right in that request.

Any help would be greatly appreciated

PR #1197, opened 2015-01-14
Filenames in ZIP not always UTF-8. Fixing this with chardet.

sourenaraya:


stephenmcd:
Can you provide a sample zip file that causes the problem?

Would rather not introduce a dependency if avoidable.

sourenaraya:
Any ZIP created by windows and contains non ascii symbols, for example cyrillic.
https://www.dropbox.com/s/13m8lj95s7fx4b5/Sample.zip?dl=0

sourenaraya:
And about dependency: encoding can be GB2312 in chinese locale, IBM866 in russian and so on. 
So, we can  keep chardet as dependency or include into mezzanine tree, not sure if possible.

stephenmcd:
Thanks for all the info - very surprising to learn about how filenames in zipfiles can use arbitrary encoding without any hints at all. I think your solution is exactly right.

Thanks a lot!

PR #1199, opened 2015-01-20
Django 1.8 compatibility

AlexHill:
Might as well get on this early. Here are the changes I needed to make to get my app to boot – haven't run the test suite yet.

There are three groups of changes:
* the updated template system, supporting multiple template engines
* further changes to the Meta API and model loading
* ``django.contrib.comments`` moved to an external package, ``django-contrib-comments``

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ env:
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.5.x.zip
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.6.x.zip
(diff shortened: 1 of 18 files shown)


AlexHill:
It's going to be a bit tricky to specify requirements if we're continuing to support 1.4, because the broken-out ``django-contrib-comments`` package specifies Django >= 1.5 in its setup.py. I suppose this could be because 1.8 is designated as an LTS release, so they felt 1.4 could be dropped. Hmm...

AlexHill:
A bunch of these commits are relevant for 1.7 support; I'll break those out into a separate pull request.

AlexHill:
Ok, all tests passing on 1.5 through to 1.8, Python 2.7, 3.3 and 3.4. Tomorrow I'll go through these changes and squash them into more sensible commits to get rid of the noise.

AlexHill:
I've split these changes into a few separate PRs since the comments issue is going to be tricky, so I'm going to close this one.

PR #1200, opened 2015-01-21
Test with Py 3.4, Django 1.7 and default runners

AlexHill:
This commit updates the Travis config to test under Python 3.4 and Django 1.7.

Though not yet necessary for 1.7, it also updates the test script to work with the old ``DjangoTestSuiteRunner`` and the new default ``DiscoverRunner`` introduced in Django 1.6. ``DjangoTestSuiteRunner`` is removed in 1.8. I've removed the line that specifies ``settings.TEST_RUNNER``, so the default runner will be used under each version of Django.

The ``DjangoTestSuiteRunner`` would run all the tests in ``django.contrib`` as well as Mezzanine's tests – ``DiscoverRunner`` doesn't do this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,19 @@ env:
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.5.x.zip
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.6.x.zip
+  - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.7.x.zip
 python:
   - "2.6"
   - "2.7"
   - "3.3"
+  - "3.4"
 matrix:
   exclude:
+     - python: "2.6"
+       env: DJANGO_VERSION=https://github.com/django/django/archive/stable/1.7.x.zip
      - python: "3.3"
+       env: DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
+     - python: "3.4"
        env: DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
 install:
   - pip install $DJANGO_VERSION --use-mirrors
--- a/mezzanine/bin/runtests.py
+++ b/mezzanine/bin/runtests.py
@@ -1,7 +1,10 @@
 from __future__ import unicode_literals
+
 import atexit
 import os
+import shutil
 import sys
+
 import django
 
 
@@ -17,6 +20,10 @@ def main(package="mezzanine"):
     os.environ["DJANGO_SETTINGS_MODULE"] = "project_template.test_settings"
     package_path = path_for_import(package)
     project_path = os.path.join(package_path, "project_template")
+
+    # Create local_settings.py so it is imported by settings.py
+    local_settings_path = os.path.join(project_path, "local_settings.py")
+    shutil.copy(local_settings_path + ".template", local_settings_path)
 
     test_settings_path = os.path.join(project_path, "test_settings.py")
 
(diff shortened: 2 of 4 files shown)


stephenmcd:
Awesome - thank you Alex!

stephenmcd:
Hey Alex - how can we maintain the previous behaviour of testing all installed apps (Django + Mezzanine) by default, even for Django 1.8? Would we need to reproduce something like DjangoTestSuiteRunner ourselves? 

Do you think this would be a good idea? What's the alternative, do we need to configure DiscoverRunner instead?

stephenmcd:
Also what's the thinking around no longer leveraging the local_settings template? Seems like we're just repeating some of its parts (sqlite db, secret/nevercache keys) in the string template we create inside runtests.py, and if that set of "default settings we absolutely need" changes over time, we'll need to update them in two places now.

AlexHill:
The ``DjangoTestSuiteRunner`` would only run tests that lived alongside the code, whereas most of Django's tests (even for contrib) lived in the ``tests`` folder, outside the ``django`` package, and so aren't even present if you install from pip or an archive. I thought that since it isn't comprehensive anyway, we might as well just focus on running our own tests.

You've reminded me that I'll need to think about Cartridge in this context though.

stephenmcd:
Thanks Alex, I didn't know that and the changes make sense now.

AlexHill:
I dropped the local_settings import at some stage because I was actually using nothing from it and it simplified the code – you're right though, if we add any new settings that need to be defined in local_settings.py then we'd need to add them in two places. I've just added it back.

Do you think we should set ``debug = False`` when running tests?

AlexHill:
I dropped the local_settings import at some stage because the only thing it was doing was adding the secret keys to the settings. You're right though, better import it to potentially avoid a gotcha down the road.

I've added it back in by making a local_settings.py from local_settings.py.template, and allowing settings.py to import it as usual. That way ``set_dynamic_settings`` is called after importing local_settings as it is in production.

stephenmcd:
> Do you think we should set DEBUG = False in tests or leave it as True as defined in local_settings?

I can't think of a good reason either way really, what do you think?

stephenmcd:
Alex I think I was a bit confusing with my comments about local_settings, sorry about that - there's a subtle difference between the old behaviour and the fix you just made. With the old behaviour, we didn't actually generate/delete a local_settings module, we just used the template file to generate the test_settings module - this meant that if you (by which I mean me) had a local_settings module already, it'd be picked up as usual via settings.py, and left intact. I just went ahead and restored the old approach, merging in your changes by hand where they made sense: https://github.com/stephenmcd/mezzanine/commit/4d9db202d9bd141bf52a70c5e66b101840be1fee

Thinking a bit about Cartridge too, I feel more like we might still need something like DjangoTestSuiteRunner - I think it's important for Cartridge to run the Mezzanine tests, having Mezzanine run at least some of Django's tests is not a bad thing. Using DiscoverRunner also means you can no longer simply run `./manage.py test` from the project_template directly, which is really handy during development (that's how I normally run the tests locally). I added a quick path hack that got that working again, which we could remove if we get something like DjangoTestSuiteRunner working again: https://github.com/stephenmcd/mezzanine/commit/3a336635ff24de2d451a5f0bf88a3da365f23ffd

Thanks again for the enormous amount of help with everything this week.

AlexHill:
OK, I think I understand what I've been missing :) To be clear – you're talking about having an existing, hand-modified local_settings.py in your project_template directory that you want to test against? I hadn't accounted for that. Presumably that's why there's the check to see if test_settings.py exists as well?

> Using DiscoverRunner also means you can no longer simply run ./manage.py test from the project_template directly

``./manage.py test ..`` should work for that.

stephenmcd:
Yeah that's basically it - it's not that I wanted to specifically test against that local_settings, it's just that it exists for development and I don't want it clobbered.

That's a nice trick passing the parent path as a arg to the test command like that, thanks.

AlexHill:
Having had a look at the DiscoverRunner source it looks like it will be pretty easy to write our own TestRunner subclass that always runs all the Mezzanine (and Cartridge if installed) tests, I'll have a crack.

I think we should decide whether or not an existing local_settings.py should be used or not when running tests. If the answer is yes (I think it should be) then we can simplify runserver.py a bit by just checking for local_settings.py, creating it from local_settings.py.template (and later deleting it) if it's not there, and simply letting settings.py import from local_settings.py as normal.

AlexHill:
To see what I mean about local_settings.py – [have a look at this one](https://github.com/AlexHill/mezzanine/blob/4ee3bc08d0ad0ecdf1e6a9b1cc60e20f5cd35caf/mezzanine/bin/runtests.py)

I think it's a good idea to simply let settings.py import local_settings when running tests, as it does when running the server.

stephenmcd:
Sorry Alex the current setup is a bit confusing, but I think it does do what you expect already. An existing local_settings.py currently *will* get imported with runtests.py - you'll see the first line in the big string we write to the generated test_settings.py is `from project_template import settings` - so that imports the normal settings.py, which includes the normal local_settings import.

PR #1202, opened 2015-01-21
Support comments under Django 1.8

AlexHill:
The comments app has been broken out into an external package in Django 1.8, available at https://github.com/django/django-contrib-comments.

However, the new standalone comments app requires Django 1.5 or above, so if we include django-contrib-comments as a dependency, we can't use Django 1.4 as it will be automatically upgraded. I've posted to the distutils-sig mailing list to see if there's any easy solution.

In the meantime, this PR will need to be modified to use whichever comments app is available instead of only using the external package as it does now. This might cause a bit of a headache with migrations, because the app name is different: ``"comments"`` in the contrib version, and ``"django_comments"`` in the standalone version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,19 @@ env:
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.5.x.zip
   - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.6.x.zip
+  - DJANGO_VERSION=https://github.com/django/django/archive/stable/1.7.x.zip
 python:
   - "2.6"
   - "2.7"
   - "3.3"
+  - "3.4"
 matrix:
   exclude:
+     - python: "2.6"
+       env: DJANGO_VERSION=https://github.com/django/django/archive/stable/1.7.x.zip
      - python: "3.3"
+       env: DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
+     - python: "3.4"
        env: DJANGO_VERSION=https://github.com/django/django/archive/stable/1.4.x.zip
 install:
   - pip install $DJANGO_VERSION --use-mirrors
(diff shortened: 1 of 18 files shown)


AlexHill:
This commit also includes the changes in #1200 and #1201.

stephenmcd:
Thanks for all your work on this Alex, it's incredibly appreciated.

At this stage it looks like Django makes it impossible to have a project like Mezzanine that supports what will be their latest release (1.8) as well as their LTS release (1.4) at the same time. That, coupled with some of the lengths you've had to go to in this work to support 1.4, I'm inclined to change our own policy to no longer mirror Django's LTS schedule, and simply go back to supporting both the latest and previous Django versions.

What do you think? Otherwise when 1.8 comes out, we'll be supporting 1.4 all the way through to 1.8 - it's a bit much.

AlexHill:
Hmm...1.4 will be supported until 6 months after the release of 1.8. I
think it's good to mirror Django's schedule – but that doesn't necessarily
mean we have to provide functional updates. Perhaps we could issue a new
release (I guess 3.3 by that time) that supports 1.8 but not 1.4 or Python
2.6, and continue to provide bugfixes to 3.2 until 1.4's EOL later this
year?

I think that would be the "best" thing to do, but it depends on how easy it
is to patch two branches and issue two releases.

Alex

stephenmcd:
I definitely agree it would be good to mirror the LTS schedule as we've been doing, but there must be some point where it becomes to much trouble, and this incompatibility with comments and versions seems like a clear case of that.

Also I feel like maintaining two Mezzanine versions would just be a different yet equal sized burden as it would be supporting all those Django versions with one Mezzanine code base.

AlexHill:
Cool – that makes this comments issue very straightforward :)

Perhaps supporting only the most recent Django-supported LTS release would
do?

Good news is we can drop 1.5 too since that's already unsupported. Might be
a good time to comb the codebase for version-specific shims and fixes and
remove any old ones.

Alex

stephenmcd:
Yep 1.5 support can definitely go. I guess we can punt on deciding what the policy is with Django LTS until Django 1.10 since we'll support 1.8 and 1.9 at least when 1.9 is out.

AlexHill:
Cool. Closing this PR in favour of one with all 1.4 and 1.5 support stripped out.

AlexHill:
Actually, might as well leave the comments changes in their own pull request.

AlexHill:
I've opened #1203 to drop 1.4 and 1.5 compatibility.

PR #1204, opened 2015-01-23
Allow 'site_id' arg when importing blog

Nick011:
- Useful for users running multiple sites from a single mezzanine
  instance to map imports from multiple blogs


stephenmcd:
Thanks Nick - there's actually already an option for this hacked into Mezzanine's manage.py script where you can pass "--site=1" into any management command to achieve the same result:

https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/project_template/manage.py#L14-L20

PR #1206, opened 2015-01-25
KeyError when excluding first_name, last_name, and username fields

rsadwick:
I was using ACCOUNTS_PROFILE_FORM_EXCLUDE_FIELDS and ACCOUNTS_NO_USERNAME settings so that I only needed an email and password.  I got snagged on a key error where save is assuming first_name and last_name are set.


stephenmcd:
Thanks Ryan

PR #1207, opened 2015-01-29
Update rating_tags.py

yuhengqiu:
Mino correction for the tag "rated", which indicates a user has rated the specific blog post. The previous code didn't function well once after the user undo a rate.


stephenmcd:
Hey there, thanks for bringing this up.

I don't think the fix is quite right for a couple of reasons - firstly you've made the functionality dependant on having an authenticated user. Secondly you've added a DB query where there previously (and importantly) were none at all - for a template that contains lots of rating forms, this could be a potentially performance blow-out. 

Wouldn't a better fix be to correctly update the cookie when a rating is undone, and would you like to take a look at that? I'll close this PR for now.

PR #1208, opened 2015-01-30
Patch 34

yuhengqiu:
Hi Stephen, you have reason. Here is the update on cookie operation. 1) set a undoRating signal; 2) clean a little bit after an auth user undoing a rate.


yuhengqiu:
Sorry, please close it, followed by patch-35

PR #1209, opened 2015-01-30
Fix the rated logic after an auth user undoing a rate.

yuhengqiu:
1) set a undoRating signal in form.py
2) reform the ratings record for set cookie


stephenmcd:
Great! Thank you.

PR #1210, opened 2015-01-31
Django rest framework

thegrymek:
added some api from rest


stephenmcd:
Hey this looks really nice and complete, thanks a lot! I'll do a thorough review when time permits.

thegrymek:
Thanks a lot. Its not complete at all :)

Currently blog posts and pages are only provided

stephenmcd:
Well that's the bulk of Mezzanine's default content really :-)

stephenmcd:
I've looked at this more closely, it needs a lot of work.

- Let's use the rest framework built-in Django filters rather than introducing a dependency.
- Pages needs a single API that deals with all page types, not an API per page type.
- What is the markdown dependency for? Please remove it.
- Tons of missing docstrings, and the docstrings that are there are poorly punctuated. Use proper casing, punctuation, etc.
- Use `published()` method on managers which needs access to the request object and user, not `.filter(status=CONTENT_STATUS_PUBLISHED)`.
- Why is there an API that publicly lists users? At best, an unauthenticated user can create a user account. A non-staff authenticated user can view/update their own record. At the least this whole API could just be removed entirely.
- URLs should be more intuitive. /api/ should return a response listing available APIs (eg https://kouio.com/api/), and things like "/api/generic/" and "/api/core/" make no sense to a consumer - these need not mimic the Django app structure.

thegrymek:
A tons of remarks! I'll look closer to them in meanwhile.

auvipy:
how about using viewsets and routers as per drf doc says?

thegrymek:
i have to read much more about viewsets and routers to use it.

* markdown removed.
* added 'rest_framework' to apps in settings.py so it will generate nice look api page. Not everytime but for particular view it works.

About Filters. DRF recommends to use 'DjangoFilterBackend' but to use it, we have to install django-filters package.
More about it:
http://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend

nym:
I was just building a basic restful api for getting blog posts and pages from a specific root page. Instead of finishing, I will try this out instead. Personally the rest_framework has a lot of great features, so I would not mind that dependency.

stevefox:
@thegrymek Looks like a great feature for Mezzanine. Message me if you want any help so we can coordinate efforts.

gcushen:
FYI, a fairly popular restful Mezzanine API package already exists on PyPi and Github - see http://gcushen.github.io/mezzanine-api/ or https://github.com/gcushen/mezzanine-api . Feel free to let me know any feature requests or if you would like to help out with development of new features.

thegrymek:
@stevefox thanks for your help. Currently I havent enough time to develop this branch. 

@gcushen As Stephen said, this api should use only and only django-rest-framework as dependency. Another think as I see is that you can list all users in one query. Summarizing, is very good package. Seems be complete.

jerivas:
Closing because a third-party alternative exists and the PR author is unable to continue development.

PR #1212, opened 2015-02-12
Delete cached context settings on settings update

dsanders11:
#### Overview

With editable settings it is best to invalidate the context settings cache key when settings have been changed via the admin panel so that they can take effect immediately.

#### Discussion

There's a change here to the name of the key under which context settings are cached. This was so that it could be deleted from the admin changelist. I don't **think** there is anything wrong with the key name change. Before it would be prefixed with the device type and the site ID, but I don't believe either of those would result in different settings values? I don't know of any way to make a setting device type or site specific. If that holds true, then this will also increase cache hits for different device types.

### DIFF
--- a/mezzanine/conf/admin.py
+++ b/mezzanine/conf/admin.py
@@ -8,6 +8,7 @@ from django.utils.encoding import force_text
 
 from mezzanine.conf.models import Setting
 from mezzanine.conf.forms import SettingsForm
+from mezzanine.utils.cache import cache_delete, cache_installed
 from mezzanine.utils.urls import admin_url
 
 
@@ -37,6 +38,8 @@ class SettingsAdmin(admin.ModelAdmin):
         if settings_form.is_valid():
             settings_form.save()
             info(request, _("Settings were successfully updated."))
+            if cache_installed():
+                cache_delete("context-settings")
             return self.changelist_redirect()
         extra_context["settings_form"] = settings_form
         extra_context["title"] = u"%s %s" % (
--- a/mezzanine/conf/admin.py
+++ b/mezzanine/conf/admin.py
@@ -8,7 +8,8 @@ from django.utils.encoding import force_text
 
 from mezzanine.conf.models import Setting
 from mezzanine.conf.forms import SettingsForm
-from mezzanine.utils.cache import cache_delete, cache_installed
+from mezzanine.utils.cache import (cache_delete, cache_installed,
+                                   cache_key_prefix)
 from mezzanine.utils.urls import admin_url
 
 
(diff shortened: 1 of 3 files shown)


dsanders11:
Not sure what is up with the CI build failure, it looks related to syntax warnings not related to my changes.

stephenmcd:
Each of the editable (eg stored in DB) settings are site specific, so this won't work.

Perhaps the deletion could iterate through all the site records and delete the cache entry for every site/device combo.

dsanders11:
Ok. Are the editable settings device specific?

If not, the prefix could be just the site ID, and when deleting the cache key it could use the current site ID.

AlexHill:
For consistency, maybe add an optional parameter ``ignore_device=False`` to ``cache_key_prefix`` and use that? Other than that, sounds like it will work.

dsanders11:
@AlexHill, good suggestion, that's a nice clean way to do it. What do you think @stephenmcd?

stephenmcd:
I feel like it'd be cleaner to just iterate through the device types when deleting.

dsanders11:
Is there any need for the device type to be part of the key prefix, though? If the settings are the same for all device types then it doesn't really need to be part of the prefix here. Just my 2 cents.

stephenmcd:
Conceptually there's no "what's the cache key prefix for settings" vs "what's the cache key prefix for foo" vs "what's the cache key prefix for bar" - there's only "what's the cache key prefix" and we use the same thing everywhere. The device portion means that when generated HTML gets cached, it'll vary per device which is what the device detection functionality is there for.

It's very consistent in as much as there only one way to do it for everything, which makes it much easier to reason about mentally than having an option passed in.

Deleting the cache entries per device in one spot means containing the ugly to one spot, versus leaking it  across the whole concept of generating cache keys, which is currently self contained.

dsanders11:
Do you have a suggestion on how to iterate the devices and generate the cache keys? I can see how to iterate the devices, but the logic to generate the cache prefix pulls the device from the request, so it seems we'd need to duplicate the code to determine the prefix.

stephenmcd:
The devices are in the setting `DEVICE_USER_AGENTS` which is a tuple of tuples (but kinda shaped like a dict where the devices would be keys) so something like:

```
for site in Site.objects.all():
    for device in settings.DEVICE_USER_AGENTS:
        cache.delete("%s-%s-context-settings" % (site.id, device(0)))
```

But you're right we need to duplicate the key logic here. Given that, I think Alex's idea would be less ugly overall. At least passing the option in would retain the logic around site and Django's cache key prefix setting, and we actually only need to delete the cache entry for the one site that's having its settings updated - I should have thought that through a bit more and had more faith in Alex too :-)

dsanders11:
Sounds like a consensus then. :-)

I'll update the branch to work this way.

dsanders11:
Is there anything more I should do here?

stephenmcd:
Don't think so David, thanks.

stephenmcd:
All merged now, thanks again David. And thanks Alex for the review, much appreciated.

PR #1216, opened 2015-02-17
Fabric enhancements

jerivas:
My overall goal with this PR is to help developers with none or minimal sys-admin knowledge get Mezzanine sites deployed following best practices. I've tested all new features and made sure existing ones were not broken.

Enhancements over the current fabfile:

- Specifically designed to host multiple sites on the same server.
- Vanilla `gunicorn` is used instead of the the deprecated `gunicorn_django`.
- You don't need to host your repos in external sites (GitHub, Bitbucket). The contents are transferred directly from your dev machine to the server.
- You can upload files to the server via `rsync` instead of `git` or `hg` (in case your project is not under VCS).
- You don't need to know which port Gunicorn is going to use, because the connection from Nginx is to a socket file.
- Operations using `sudo` only require you to type the password once.
- Invalid requests (hosts other than `ALLOWED_HOSTS`) are blocked on Nginx level.
- Static files are set to expire after 30 days in browser cache (best practice).
- Deployment settings have been moved to `local_settings.py` for security concerns (they shouldn't be tracked by VCS) and to make sure `SECRET_KEY` and `NEVERCACHE_KEY` are in scope.
- Mezzanine sites in the server automatically start, even after system reboot.
- Locales are generated for every new site, instead of once per server.
- Added a bunch of comments and updated docstrings to make the whole fabfile more "didactic".

I've written a quick HOWTO (which I recommend is included in the official docs if this gets merged in): https://gist.github.com/jerivas/c36c41d25fc78614b18c. I would like to do a screencast to go along with that.

I have some left over credit on my Digital Ocean account in case Steve or anybody wants to test this out in a real VPS. As far as my own testing goes, I've been using Ubuntu 14.04 and tried out all three deployment methods (git, mercurial and rsync, including the `rollback` and `remove` tasks on each of them).

### DIFF
--- a/mezzanine/project_template/deploy/gunicorn.conf.py.template
+++ b/mezzanine/project_template/deploy/gunicorn.conf.py.template
@@ -1,7 +1,8 @@
 from __future__ import unicode_literals
 import multiprocessing
 
-bind = "127.0.0.1:%(gunicorn_port)s"
-workers = multiprocessing.cpu_count() * 2 + 1
+bind = "unix:%(proj_path)s/gunicorn.sock"
+workers = %(num_workers)s
+errorlog = "/home/%(user)s/logs/%(proj_name)s_error.log"
 loglevel = "error"
 proc_name = "%(proj_name)s"
--- a/mezzanine/project_template/fabfile.py
+++ b/mezzanine/project_template/fabfile.py
@@ -1,5 +1,5 @@
 from __future__ import print_function, unicode_literals
-from future.builtins import input, open
+from future.builtins import open
 
 import os
 import re
@@ -10,9 +10,14 @@ from glob import glob
 from contextlib import contextmanager
 from posixpath import join
 
-from fabric.api import env, cd, prefix, sudo as _sudo, run as _run, hide, task
+from fabric.api import abort, env, cd, prefix, sudo as _sudo, run as _run, \
+    hide, task, local
+from fabric.context_managers import settings as fab_settings
+from fabric.contrib.console import confirm
 from fabric.contrib.files import exists, upload_template
+from fabric.contrib.project import rsync_project
 from fabric.colors import yellow, green, blue, red
+from fabric.decorators import hosts
 
 
 ################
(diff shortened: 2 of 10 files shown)


AlexHill:
Though I don't use the fabfile (I like uWSGI) it's fantastic to see this getting attention. Good job Eduardo.

mezzanine/project_template/deploy/gunicorn.conf.py.template
```diff
@@ -1,7 +1,8 @@
 from __future__ import unicode_literals
 import multiprocessing
 
-bind = "127.0.0.1:%(gunicorn_port)s"
-workers = multiprocessing.cpu_count() * 2 + 1
+bind = "unix:%(proj_path)s/gunicorn.sock"
+workers = 2
```
stephenmcd: Why is this hard-coded? I think the common case would be to leverage the number of cores on the machine. Perhaps we can have it default to that, and overideable via a setting.

mezzanine/project_template/deploy/local_settings.py.template
```diff
@@ -15,7 +15,7 @@ DATABASES = {
         # Not used with sqlite3.
         "PASSWORD": "%(db_pass)s",
         # Set to empty string for localhost. Not used with sqlite3.
-        "HOST": "127.0.0.1",
+        "HOST": "localhost",
```
stephenmcd: Why this change? It's not guaranteed localhost exists (will probably never happen, but there's at least one reason for having it one way over the other).

mezzanine/project_template/deploy/supervisord.conf
```diff
@@ -0,0 +1,18 @@
+[unix_http_server]
```
stephenmcd: Why provide this file at all? What benefit over the distribution's defaults?

mezzanine/project_template/fabfile.py
```diff
@@ -333,6 +388,29 @@ def manage(command):
     return run("%s %s" % (env.manage, command))
 
 
+###########################
+# Security best practices #
+###########################
+
+@task
+@log_call
+@hosts(["root@%s" % host for host in env.hosts])
```
stephenmcd: Does this mean a root account is required? Is that assuming too much?

stephenmcd:
Looks good overall, I just left some questions inline.

Once we're happy with everything, can we squash this all into one commit? All the commits listed above will end up in the changelog otherwise.

Thanks a lot Ed!

jerivas:
Thanks for the feedback, Steve. I'll go over all your points and make some
tweaks later this week (hopefully).

mezzanine/project_template/fabfile.py
jerivas: For the `secure` task: yes, we're assuming root access. The idea is you only need this task when you receive a brand new VPS (which means you get the root password in an email). If you already have a VPS with a regular user with sudo privileges, you can skip this task completely and use the others with said user. If you are in a shared hosting environment where sudo is not available, then this fabfile is not really for you :P

mezzanine/project_template/deploy/gunicorn.conf.py.template
jerivas: In my experience you usually have access to 8 or 16 cores but as little as 512Mb or 1Gb of RAM in a VPS. When running multiple sites you end up with a bunch of workers that eat up your memory. However, single sites probably don't have this problem. I'll add the setting as you suggest as it seems a nice compromise.

mezzanine/project_template/deploy/local_settings.py.template
jerivas: I don't remember exactly why I went with `localhost` (it was related to some DB connection issues), but I'll test it and change it back to `127.0.0.1` if I can't reproduce the problem.

mezzanine/project_template/deploy/supervisord.conf
jerivas: Mostly to be able to run `supervisord` and `supervisorctl` as the current user instead of requiring sudo (as a convenience when doing server maintenance). However, I just brought this over from the alternate fabfile and will further explore using the distro default to get rid of this file.

jerivas:
Oh, and yes, I'll squash the commits into one when we're done. What would you like it to say?

stephenmcd:
Thanks for all the answers Ed, they all make sense. As for the commit message, just a brief description of the enhancements I guess.

jerivas:
Okay, I addressed all concerns and ended up removing a good deal of unnecessary changes/additions, so thanks for the sharp eye, Steve. Let me know if we're good to go and I'll squash all the commits.

jerivas:
Also, let me know if you'd like me to update the docs. As I see it, all that needs to be done is update the settings code block and Fabric tasks descriptions. I would also like to propose we include a mini "tutorial", because the current reference might not be enough for complete newcomers. I was thinking on something like [this](https://gist.github.com/jerivas/c36c41d25fc78614b18c).

stephenmcd:
Thanks Ed. The mini tutorial is a great idea - many people have complained about the lack of docs on using the built-in fabfile.

mezzanine/project_template/fabfile.py
```diff
@@ -40,24 +45,31 @@
...
 env.domains_python = ", ".join(["'%s'" % s for s in env.domains])
 env.ssl_disabled = "#" if len(env.domains) > 1 else ""
-env.repo_url = conf.get("REPO_URL", "")
-env.git = env.repo_url.startswith("git") or env.repo_url.endswith(".git")
+env.vcs_tools = ["git", "hg"]
+env.deploy_tool = conf.get("DEPLOY_TOOL", "rsync")
 env.reqs_path = conf.get("REQUIREMENTS_PATH", None)
-env.gunicorn_port = conf.get("GUNICORN_PORT", 8000)
 env.locale = conf.get("LOCALE", "en_US.UTF-8")
+env.num_workers = conf.get("NUM_WORKERS",
```
jerivas: If a user doesn't specify a number in this setting, it'll default to the recommended `cores*2+1`. You can see it gets injected in the right place in gunicorn_conf.py.template.

jerivas:
Ok I squashed the commits and left only two: one mentioning the fabfile changes and the other one mentioning the doc changes. I had to pull from master and I'm not sure of how to squash that last commit, so any pointers are appreciated.

stephenmcd:
Thanks Ed!

I use mercurial for everything (which gets mirrored via git on here) so I can't really help there.

One extraneous commit message isn't too big a deal though.

fedelibre:
I think you should rebase master in your branch to have your commits on top
and then:

git commit --amend

To add your last change to the last commit

jerivas:
Thanks Federico. Rebasing instead of merging was the solution. Now we have a squeaky clean commit history.

stephenmcd:
All merged - thanks again Ed for the huge effort.

PR #1217, opened 2015-02-18
Pep8 fixes

Kniyl:
Account for the recent Travis-CI build configuration changes and remove pep8-based fails.


Kniyl:
…and the build fails…
It should be fine, it's just a pip error for one build. Every other works fine.

stephenmcd:
Thanks Mathias - I opted for a simpler cleanup by allowing us to keep using lambdas: b0470f9c1b63a18913adf7cf8f2ea335385efe2c

PR #1219, opened 2015-02-25
Don't assume INSTALLED_APPS is all modules as Django 1.7 adds AppConfig

frankier:
As in https://github.com/stephenmcd/mezzanine/pull/941 - we should just ignore INSTALLED_APPS which are appconfig classes to avoid crashes.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -49,6 +49,13 @@ page, we would define the following in ``authors.defaults``::
         default=10,
     )
 
+.. note::
+
+    If you are using Django 1.7 or greater and your app is included in your
+    INSTALLED_APPS as an AppConfig (eg authors.apps.MyCrazyConfig), Mezzanine
+    won't import your defaults.py automatically. Instead you must import it
+    manually in your AppConfig's ready() method.
+
 Reading Settings
 ================
 
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -200,11 +200,15 @@ class Settings(object):
 
 mezz_first = lambda app: not app.startswith("mezzanine.")
 for app in sorted(django_settings.INSTALLED_APPS, key=mezz_first):
-    module = import_module(app)
     try:
-        import_module("%s.defaults" % app)
-    except:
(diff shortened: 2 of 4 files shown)


frankier:
The broken Travis build seems to be a result of PEP8 style issues currently in HEAD.

stephenmcd:
I don't understand the change. 

Wouldn't something like `get_app_name_list()` as per #941 be needed?

frankier:
It's not needed unless we want to support page_processors for AppConfig apps, but perhaps a little better. Note it's safe to assume that by the time this code runs everything in INSTALLED_APPS is either a module or an AppConfig. Django would've already bailed out if there were another problem. This means errors aren't being swallowed and Mezzanine works as well as before, simply it doesn't crash when used in a project that specifies AppConfigs in INSTALLED_APPS.

I've rebased & added `get_app_name_list()`

stephenmcd:
This is really confusing.

Doesn't the change in the `conf` app mean that some apps will be ignored, and that their `defaults.py` modules might not be picked up and registered? That's what my original question was with regard to, and I was asking whether the approach in #941 is what's actually needed in the `conf` app. But now you've brought in the changes from #941, and the changes to the `conf` app are still the same?

frankier:
Yes. It will ignore apps using explicit AppConfigs. Note we're not losing anything here since currently an explicit AppConfig will cause an exception to reach the top of the stack/crash. No, we can't use get_app_name_list() for defaults because the code at the top level of mezzanine/conf/__init__.py is run before the AppConfig initialisation phase.

This would be thrown:
https://github.com/django/django/blob/9a4a9a8a49ed3e38f3038b8d6fc7c405256271f1/django/apps/registry.py#L121

People using an explicit AppConfig can import the defaults themselves in the ready method of their AppConfig. I have added this to the docs. As far as I can see this is the most reasonable option since there doesn't seem to be a way to get a signal after the app registry is initialised (if there were we could do it then for Django >= 1.7 instead).

stephenmcd:
Ok that's a bit clearer, thanks a lot.

PR #1221, opened 2015-03-01
fix transifex url on README.rst

luzfcb:
changed `transifex.net` to `transifex.com` because `transifex.net` has an expired SSL certificate also because the official address is `transifex.com`


stephenmcd:
Thanks Fábio

PR #1224, opened 2015-03-05
Handle new app-loading mechanism for Django 1.7 in fabfile.

jerivas:
See https://docs.djangoproject.com/en/1.7/releases/1.7/#standalone-scripts
Django 1.7 requires explicitly calling `django.setup()` when using Django outside management commands. I've added an inline "if" statement to support Django 1.6 and 1.7. In my tests this was the only issue with Django 1.7 when deploying. Perhaps the next step is to not automatically bundle South when installing the server reqs.


stephenmcd:
Nice one, thanks Ed.

Issue #1228, opened 2015-03-12
core.templatetags.thumbnail can throw unhandled exception


abendig:
In `core.templatetags.thumbnail`, at line 307, we find the following code:
```
if not os.path.exists(thumb_dir):
    os.makedirs(thumb_dir)
thumb_path = os.path.join(thumb_dir, thumb_name)
```

The `os.makedirs` may [raise an exception](https://docs.python.org/2/library/os.html#os.makedirs), which this code is not accounting for. Yes, there is a check for whether the path already exists (one potential reason for an exception), but the exception can still be raised if the method fails to create the path for some other reason.

I ran into this, as I was debugging issues related to using a custom storage backend.

One way to reproduce the exception is by attempting to write to a read-only directory:

```
>>> import os
>>> os.mkdir("makedirtest", 0544)
>>> os.makedirs("makedirtest/test")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/anaconda/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: 'makedirtest/test'
```
Particularly, since `thumbnail()` is called from template code, it should be handling the exception somehow. Happy to provide a fix; I think even a minimal try/except block may be sufficient here. Thoughts?

stephenmcd:
Yep let's just wrap the dir creation in try/catch for appropriate exception.

Thanks a lot for your help!

stephenmcd:
#1229

PR #1229, opened 2015-03-12
Added exception handler for make_dirs() function call in thumbnail() template tag

abendig:
This change is to deal with `make_dirs` raising exceptions, such as when attempting to write to a read-only directory or when working within an environment with customized Python `os` module, as is the case on Google App Engine.

This is for issue https://github.com/stephenmcd/mezzanine/issues/1228.


stephenmcd:
Awesome, thanks again!

Issue #1148, opened 2014-11-05
Improve API Documentation


prikhi:
I would love it if the docs took advantage of some more of Sphinx's features, like cross-references and info field lists: http://sphinx-doc.org/domains.html#the-python-domain

I'm willing to get started on this, I just want to get the sign off before putting any work into it.
Would you prefer a single(squashed?) PR or a PR per app?

And is there opposition to enabling the `sphinx.ext.viewcode` Sphinx extension? I find this helpful as it keeps me from having to jump between mezzanines docs when I need the overview and the code when I need the details. But I wouldn't be sad if this got shot down, I can always just build them locally...
## mezzanine/
- [ ] accounts
- [ ] blog
- [ ] boot
- [ ] conf
- [ ] core
- [ ] forms
- [ ] galleries
- [ ] generic
- [ ] mobile
- [ ] pages
- [ ] template
- [ ] twitter
- [ ] utils
## docs/
- [ ] admin-customization
- [ ] blog-importing
- [ ] caching-strategy
- [ ] colophon
- [ ] configuration
- [ ] content-architecture
- [ ] deplyment
- [ ] device-handling
- [ ] fabfile
- [ ] frequently-asked-questions
- [ ] index
- [ ] inline-editing
- [ ] model-customization
- [ ] model-graph
- [ ] multi-tenancy
- [ ] overview
- [ ] packages
- [ ] search-engine
- [ ] settings
- [ ] twitter-integration
- [ ] user-accounts
- [ ] utilities

This is a lot and my mezzanine dev time is limited so I will be slowly working on this over at prikhi/mezzanine@c19b41b73559586a7894a7b9fa5e800b6cdfff34

auvipy:
@prikhi looking forward to it. any update?

prikhi:
@auvipy I've found myself infinitely busy(as usual). I could quickly add `sphinx.ext.viewcode`. Since my time is limited, I'll probably do a PR per app, otherwise any progress will take forever to get merged. 

If you have a desire to contribute, feel free to submit a PR & reference this issue. I will check off whatever you get in.

auvipy:
I'm not good in this. can you provide any way following which i can help in this process? @prikhi

jerivas:
[...]

PR #1233, opened 2015-03-23
Show Links to Code in Package Documentation

prikhi:
Refs #1148

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,7 +51,7 @@ if sys.argv[-2:] == ["docs", "docs/build"]:
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ["sphinx.ext.autodoc"]
+extensions = ["sphinx.ext.autodoc", "sphinx.ext.viewcode"]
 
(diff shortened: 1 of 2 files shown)


stephenmcd:
Awesome, thanks so much!

stephenmcd:
Merging this, but I can't actually see what affect it has - any idea?

prikhi:
It adds the `[source]` links on the right here:
http://orotau-docs-mezzanine.readthedocs.org/en/latest/packages.html#module-mezzanine.generic

So you can quickly switch between the API docs & the implementation.

prikhi:
And some day, I will add cross-references to the non-autodoc stuff, so from sections like [Keywords](http://orotau-docs-mezzanine.readthedocs.org/en/latest/utilities.html#keywords) you can reach the [keywords_for](http://orotau-docs-mezzanine.readthedocs.org/en/latest/packages.html#mezzanine.generic.templatetags.keyword_tags.keywords_for) autodoc, and then from the `[source]` links you can reach the actual code for the `keywords_for` function. All without leaving the comfort of you browser.

stephenmcd:
Ah that's cool, when I built the docs locally I don't see that - I might need to update my version of sphinx, not sure.

Thanks

PR #1234, opened 2015-03-24
Improve modeltranslation integration for cartridge support

Kniyl:
Some bugs were left in the js of the previous pull request and cartridge can no longer override mezzanine.pages templates. So some adjustments had to be made.


stephenmcd:
Thanks so much!

PR #1238, opened 2015-03-29
Import WordPress draft posts correctly

jvarho:
Currently import_wordpress fails if there are any draft posts due to missing pub_date, as discussed e.g. [here][1]. This patch makes them import correctly as drafts.

Publish date gets set as the current date, which might not be optimal, but changing that would require delving too deep into the code and this seems to work OK.

[1]: https://groups.google.com/forum/#!topic/mezzanine-users/P9NnwfAjvKs


stephenmcd:
Fantastic - thanks a lot Jan.

PR #1239, opened 2015-03-30
Support max width/height for thumbnails

simenheg:
Support setting a maximum width and height in the thumbnail template
tag. The extra arguments `max_width` and `max_height` are introduced for
this. Contrary to the existing `width` and `height` arguments, these
won't force the thumbnail to a given size, but will only shrink it if
the image is larger than the specified limits.

I did this as a quick fix for myself because I needed this functionality, but I
figured someone else might find this useful to have as well. :)

### DIFF
--- a/docs/utilities.rst
+++ b/docs/utilities.rst
@@ -261,11 +261,14 @@ height::
 The ``thumbnail`` template tag also accepts several other optional
 arguments for controlling the generated thumbnail:
 
+  * ``no_grow`` - A boolean controlling whether the thumbnail is
+    forbidden to grow beyond its original size when resizing (defaults
+    to False)
   * ``quality`` - A value from 0 to 100 controlling the JPG quality
     (defaults to 95)
   * ``left`` and ``top`` - Values from 0 to 1 controlling where the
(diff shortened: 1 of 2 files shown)


stephenmcd:
Thanks for this - I think it's a good idea, but I'm not sure about the implementation. 

For example why is it possible to specify both `width` and `max_width` at the same time? Would it make more sense to just have `width` and `height` args, and an extra bool arg that determines whether the sizes are either forced to shrink/grow the image, or only shrink (as the intention with `max_width` and `max_height`)?

stephenmcd:
Will also need relevant documentation updated: https://github.com/stephenmcd/mezzanine/blob/master/docs/utilities.rst#thumbnail

simenheg:
Thanks for the feedback. I've pushed something I think resembles the implementation you described, together with the updated documentation. See if you find that to be closer to what you'd like. :)

stephenmcd:
Looks good! I might change the arg name, "no_grow=False" is kind of a double-negative, but leave that to me.

Thanks a lot!

PR #1243, opened 2015-04-03
Don't consider proxy models a content model.

gavinwahl:
Otherwise we'll try to use them with select_related, which is not supported.


stephenmcd:
Thanks Gavin

PR #1245, opened 2015-04-04
German translation for search-form added

christianwgd:


christianwgd:
Added the translations for search-form

PR #1247, opened 2015-04-06
Fix createdb to work with Django 1.7 and 1.8.

jerivas:
[Django 1.8 changed the way optional args are added to management commands](https://docs.djangoproject.com/en/1.8/howto/custom-management-commands/#accepting-optional-arguments). This patch adds optional args "the old way" for Django 1.7 in `Command.__init__()`, and then uses "the new way" for Django 1.8, which is via a call to the new class method `add_arguments()`.

There is a downside. In Django 1.7 we can get all args for `migrate` by simply extending `migrate.Command.option_list`. In Django 1.8, this returns an empty tuple, and I have no idea how to actually get the args from the `migrate` command, so I manually implemented `--noinput` and `--nodata`. This means in the end that `createdb` has less command line options available in Django 1.8 than in 1.7. If you want to have all options available, they are available [here](https://github.com/django/django/blob/master/django/core/management/commands/migrate.py#L27-L48) and can be directly copy-pasted. I do consider that `--noinput` and `--nodata` are good enough for `createdb`, any other task can call `migrate` directly.

### DIFF
--- a/mezzanine/core/management/commands/createdb.py
+++ b/mezzanine/core/management/commands/createdb.py
@@ -4,16 +4,13 @@ from future.builtins import int, input
 from optparse import make_option
 from socket import gethostname
 
-from django.core.management.base import NoArgsCommand, CommandError
+from django import VERSION
+from django.core.management.base import BaseCommand, CommandError
 from django.contrib.auth import get_user_model
 from django.contrib.sites.models import Site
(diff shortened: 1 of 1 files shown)


stephenmcd:
Looks good enough to me Ed, thanks a lot!

PR #1251, opened 2015-04-07
Add SECRET_KEY to docs config which fixes broken rtd builds

sjkingo:
Looking at the build history on RTD, [it hasn't successfully built the docs](https://readthedocs.org/builds/mezzanine/2541300/) in over 3 months. This is because of the change to `SECRET_KEY` in Django, and that setting not being available during the tests.

This patch changes the docs configuration to import a special settings file that provides a random `SECRET_KEY`, so Django's `setup()` can complete successfully.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,10 +20,9 @@ import mezzanine
 
 if "DJANGO_SETTINGS_MODULE" not in os.environ:
     docs_path = os.getcwd()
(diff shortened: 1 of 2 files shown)


stephenmcd:
Nice one, thanks Sam.

sjkingo:
Doh.. didn't look carefully enough before I created this and instead broke it further. See PR #1252 for the fix.

PR #1253, opened 2015-04-07
Attempt to fix path issue in rtd environment

sjkingo:
Seems this is [still broken on rtd](https://readthedocs.org/builds/mezzanine/2541862/).. it works fine with the local Sphinx builder but I'm guessing the environment they use is different.

I believe this should fix it. I've also re-factored a little.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,10 +20,10 @@ import mezzanine
 
 if "DJANGO_SETTINGS_MODULE" not in os.environ:
     docs_path = os.getcwd()
-    parts = (docs_path, "..")
-    sys.path.insert(0, os.path.realpath(os.path.join(*parts)))
-    settings_module = "docs_settings"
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -20,10 +20,10 @@
 
 if "DJANGO_SETTINGS_MODULE" not in os.environ:
     docs_path = os.getcwd()
-    parts = (docs_path, "..")
-    sys.path.insert(0, os.path.realpath(os.path.join(*parts)))
-    settings_module = "docs_settings"
-    os.environ["DJANGO_SETTINGS_MODULE"] = settings_module
+    mezzanine_path_parts = (docs_path, "..")
+    sys.path.insert(0, docs_path)
```
sjkingo: This is the line that should fix the broken build environment.

PR #1259, opened 2015-04-09
Fixes #1258: fix overextends tag for Django 1.7 & 1.8

### LINKED ISSUE #1258: overextends tag broken in Django 1.7+1.8
Looks like the changes made to ``loader_tags.py`` in a50de50699bb6a24bfb5f118449991aa7608b426 either didn't work or both Django versions have since changed.

As reported here: https://groups.google.com/d/msg/mezzanine-users/_QWfFVB3RVc/ZirizEV9t2YJ

Just pinging @AlexHill as you might have a head's up on this one already. 

I made a quick attempt by changing ``find_template_loader = context.engine.find_template_loader`` to ``find_template_loader = context.engine.find_template_loader`` which appears to work for 1.8, but then other possibly unrelated exceptions came up.

BTW my quick tip for actually running ``overextends`` is to modify the first line of `core/templates/admin/base_site.html` to use it instead of ``extends``

AlexHill:
See #1258 for discussion.

### DIFF
--- a/mezzanine/template/loader_tags.py
+++ b/mezzanine/template/loader_tags.py
@@ -49,17 +49,18 @@ class OverExtendsNode(ExtendsNode):
         import django.template.loaders.app_directories as app_directories
         try:
             # Django >= 1.8
-            app_template_dirs = app_directories.get_app_template_dirs
+            get_app_template_dirs = app_directories.get_app_template_dirs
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks again

AlexHill:
No probs!

PR #1260, opened 2015-04-09
Update Spanish translation

jerivas:
Just brought up all Spanish strings up to date and changed a few words here and there for consistency. Also caught that the `ADMIN_MENU_COLLAPSED` setting wasn't translatable and fixed that.


stephenmcd:
Awesome, thanks a lot Ed.

PR #1270, opened 2015-04-16
Change redirect to be permanent

radius:
The redirect middleware is currently using HttpResponseRedirect, which responds with http status 302.
Django does permanent redirects using HttpResponsePermanentRedirect, which sends a 301 status.

This change makes sense because:

1. This makes the Mezzanine redirection middleware more closely resemble Django
2. I guess this is more of an opinion than a fact, but it's worth mentioning that most changes on CMS-driven sites tend to be because of a re-organization, where path changes are understood to be permanent (301) rather than temporary (302).


stephenmcd:
Makes sense, thanks a lot!

PR #1282, opened 2015-05-06
Update setup.py

sanjivj:
Removed the greater-than sign from Line 54; having it say "django >= 1.7" caused the package to install django 1.8.2, which breaks the functionality.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ try:
         packages=find_packages(),
         install_requires=[
             "django-contrib-comments",
-            "django >= 1.7",
+            "django = 1.7",
             "filebrowser_safe >= 0.3.4",
             "grappelli_safe >= 0.3.12",
             "tzlocal >= 1.0",


AlexHill:
Hi @sanjivj,

The current master supports Django 1.8, so the ``>= 1.7`` is correct.

Can you post some details of the specific problem you're having running Mezzanine under 1.8?

Alex

stephenmcd:
Closing this since the change is invalid, but as Alex said please provide some detail.

sanjivj:
Stephen,

I re-examined the problem; I acknowledge that Mezzanine/Cartridge support
Django-1.8. However, many third-party plugins do not, since 1.8 is fairly
new. Presumably, that is why you still support django-1.7, since, for a
while, some developers will go on using plugins that are not 1.8-ready.

The problem is that while installing Cartridge, at one point there seems to
be a Mezzanine requirement that a django-1.6 version is installed:

Collecting django!=1.6.0,<1.7,>=1.4.10 (from mezzanine>=3.1->Cartridge)

At this point, django-1.6.11 is installed, after uninstalling django-1.7.7.
Further along, there is the requirement that django >= 1.7. Since the
existing version of django is then 1.6.11, this django version is
uninstalled and django-1.8.1 is installed!

I acknowledge that it might be impractical or imprudent to change the
behavior that is causing this, but perhaps developers could be warned that
they will have to manually re-install django-1.7 (if needed). Hope this
helps.

- Sanjiv

AlexHill:
Hey @sanjivj,

Thanks for the feedback. If Mezzanine's development version is already installed in your environment when you install Cartridge, pip shouldn't try to install dependencies for an older version, which is what appears to be happening here.

I've added a note to the docs suggesting this: https://github.com/stephenmcd/cartridge/commit/a81fc1d19cc89cf04c1445a45a9b1a363666380c

Cheers,
Alex

PR #1283, opened 2015-05-06
Add the featured_image as an enclosure to the rss feed

arcticlinux:
Add the featured image to the rss feed as an enclosure url, so that it can be used by feed readers
For example Zapier RSS to Facebook to use as the Post Image

### DIFF
--- a/mezzanine/blog/feeds.py
+++ b/mezzanine/blog/feeds.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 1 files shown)


stephenmcd:
Really nice, thanks a lot!

arcticlinux:
Not 100% sure the best way to generate the URL in that situation, I didn't think or couldn't quickly find request being available

stephenmcd:
I'm actually trying to work this out now - it's a bit of a rabbit hole.

Leave it with me and thanks again!

PR #1284, opened 2015-05-07
Adding 4 additional sites

joshbatchelor:


stephenmcd:
Wow these all look really slick.

Thanks Josh.

stephenmcd:
BTW since all the sites from your company clearly have had a professional designer work on them, feel free to always include them as "featured sites" simply by including a trailing slash in their URLs. I don't think there's been a single site from your company that hasn't been featured that way since they all look slick.

Thanks again!

stephenmcd:
Eg: 2f37a16f4064662284dfce482035d55f6945bb2f

Issue #1275, opened 2015-04-22
Activating multilanguage feature. Error '_fill_fields_cache'


itztli:
I am following the new feature of mezzanine for multilanguage support. I download the latest version from git and made a project form scratch in the following way:

1) mezzanine-project mlt

2) Configure my database settings and run:  
python manage.py createdb

3) Add in settings.py:
USE_MODELTRANSLATION = True
MODELTRANSLATION_DEFAULT_LANGUAGE = 'es'
LANGUAGE_CODE = "es"
LANGUAGES = (
    ('en', _('English')),
    ('es', _('Spanish')),
)

4) Run: 
python manage.py sync_translation_fields

and I got the error:
Traceback (most recent call last):
  File "manage.py", line 34, in <module>
    execute_from_command_line(sys.argv)
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/Django-1.8-py2.7.egg/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/Django-1.8-py2.7.egg/django/core/management/__init__.py", line 312, in execute
    django.setup()
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/Django-1.8-py2.7.egg/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/Django-1.8-py2.7.egg/django/apps/registry.py", line 115, in populate
    app_config.ready()
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/modeltranslation/apps.py", line 11, in ready
    handle_translation_registrations()
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/modeltranslation/models.py", line 81, in handle_translation_registrations
    autodiscover()
  File "/home/vdelaluz/virtualenvs/multi-language-test/local/lib/python2.7/site-packages/modeltranslation/models.py", line 32, in autodiscover
    import_module(module)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
[...]

PR #1291, opened 2015-05-13
Ensure consistent related_model attribute throughout django fields

Kniyl:
Fixes an issue from #1275

Added an exception in case some projects created classes derived from ```BaseGenericRelation``` to warn them about the change.


stephenmcd:
Congratulations on working with one of the hairiest parts of the project :-)

Looks good, I'll just test it out and merge if I don't find any problems.

Thanks Mathias.

PR #1296, opened 2015-05-17
Switch forms admin to use the object-tools-items block in template

dsanders11:
Can't currently test this, but it's nothing too complicated.


stephenmcd:
Seems to work, thanks David. 

BTW can't you test it by installing the dev branch of grappelli-safe?

dsanders11:
I was pressed for time earlier when getting that done, so that's what I meant by unable to test. :-)

stephenmcd:
Ah ok! No worries.

PR #1297, opened 2015-05-17
Re-activate modeltranslation tabs for settings in admin

Kniyl:
They were gone in @2486606838c16464340ccf68a69ef421c0e6a3dd

### DIFF
--- a/mezzanine/conf/admin.py
+++ b/mezzanine/conf/admin.py
@@ -9,6 +9,7 @@ from django.utils.translation import ugettext_lazy as _
 from django.utils.encoding import force_text
 
 from mezzanine.core.admin import BaseTranslationModelAdmin
+from mezzanine.conf import settings
 from mezzanine.conf.models import Setting
 from mezzanine.conf.forms import SettingsForm
 from mezzanine.utils.cache import (cache_delete, cache_installed,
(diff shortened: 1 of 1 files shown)


stephenmcd:
Strange, I think the intention there was that they'd show up via inheriting `BaseTranslationModelAdmin` - does that not work?

Kniyl:
JQuery ui is not loading without explicitly adding it in the Media.

Plus there are two different js files for modeltranslation tabs. One for the pages and one for the settings.

stephenmcd:
Ah I see now! So how about we do something with the `js` property, similar to what I did with the `css` property - copy its contents then modify it. That way if we need to change the base set of files we only do it in one spot.

What do you think?

stephenmcd:
Something like this? 

```
    class Media(BaseTranslationModelAdmin.Media):
        css = copy(BaseTranslationModelAdmin.Media.css)
        css["all"] += ('mezzanine/css/admin/settings.css',)
        js = [js.replace("tabbed_translation_fields.js",
                         "tabbed_translation_settings.js")
              for js in BaseTranslationModelAdmin.Media.js]
```

stephenmcd:
It's a bit hacky but at least we get some kind of re-use happening.

Kniyl:
I was messing around with something like
```python
    js = copy(BaseTranslationModelAdmin.Media.js)
    if len(js):
        js = js[:2] + ("mezzanine/js/admin/tabbed_translation_settings.js",)
```
but it feels even worse.

stephenmcd:
Yeah, it's a bit more fragile since it depends on the length + order.

I've got the above code working, I'll commit that. 

Sorry for breaking it!

Kniyl:
No problem :)

Kniyl:
Beware, though. I wasn't very consistent with files names.

It's ```tabbed_translatable_settings``` and not ```tabbed_translation_settings```

stephenmcd:
Yeah I missed that at first then saw it when testing :-)

Here's my fix: f5ea57831bae8631e9a7216e9d2cf1bcff17da20

Thanks for picking this up.

PR #1298, opened 2015-05-18
Added SITE_LOGO as setting which shows site logo 

b3h74d:
Added SITE_LOGO as setting which shows site logo instead of Site Title, adding Site Title as ALT text.


jerivas:
Hi, B3.

I'm not sure about this. IMO, the way the logo is implemented should be left to the developer. In your project you can have your own app with a `defaults.py` to achieve exactly this. But that's just me, let's see what others have to say.

Oh, by the way, the Travis build is failing because you have some PEP8 errors.

philipsouthwell:
Hi there,

Good to meet you both. From the perspective of someone reasonably new to the django/mezzanine ecosystem I think this would be a positive move forward. It's a feature which almost every site needs and would place one less barrier in the way of new people using Mezzanine. Those who wanted to deal with the logo differently could still do so within their own app.

Coincidentally last night I started to create my first reusable template as an app and this is the first thing I'm trying to do. I wouldn't have known even where to start for such an important feature when I was just starting out!

b3h74d:
Hi,

Thank you for the comments, I am also new to django/mezzanine. As @philipsouthwell mentioned it's a feature which every site needs, so I think this can help a lot to spend time better for new comers.

stephenmcd:
Yeah I agree with Ed - this goes beyond default templating and should be left up to the developer.

On the other hand, having the Mezzanine logo appear in the default templates does have some branding benefits for the project, so I don't mind it so much either.

Any other thoughts?

Kniyl:
For a personal approach, I don't mind that much since I (almost) always override ```base.html``` in my projects.

But I think it can help in various ways. For newcommers and learners as stated by @philipsouthwell or @b3h74d. Theming community could also benefit from this simpler approach of personnalizing a custom theme. I mean… it's not like a custom theme could not be shipped with its own ```defaults.py``` but people tend to stick to what is officialy supported, at least on their first try.

It's one of these features that goes “meh” for someone used to the project but can help more people getting into it. Plus, its harmless as it degrades to the actual template if the setting is not provided.

kenbolton:
My concern for developers new to Mezzanine is the nearly-overwhelming number of editable settings. This is particularly difficult for non-technical content managers, whose eyes I have observed glazing over when they go to /admin/conf/setting/. I would prefer to coach developers to explore the "register_setting" method and edit templates rather than spoon-feeding yet-another-editable-setting.

joshcartme:
I really like the idea of making the logo user editable, but I think storing the path to it in the settings is a bit brittle.  I think it's easy for a developer/computer savvy person, but I think finding the path to an image may go over the heads of a lot of admin users.

My preferred method for making content (like a logo) backend editable is this, http://bitofpixels.com/blog/on-singletonadmins-and-sitewidecontent-editing-sitewide-content-in-mezzanines-admin/

Basically a model that is a subclass of SiteRelated (let's call it Theme) and is registered with the admin using the SingletonAdmin class.  The benefits I see are:

1.  Familiar format to admin users
2.  Familiar widgets for admin users (i.e. the logo can use a FileField and have the nice image selector widget)
3.  The editable template tag works (afaik you can't use editable with settings)
4.  Like Ken said, the settings can be overwhelming to the uninitiated.  I find it helpful to keep logic in the settings, and content in my Theme model.  There is probably some overlap, but I think it's helpful to keep them as separate as possible

I think something like what I outline in that blog post could be worth including in Mezzanine.  We could have an AbstractTheme model that just included the logo (maybe title and tagline) and then a default Theme that just inherited from that.  Developers could override the theme, inheriting from the base model if they wanted.  Instead of using a templatetag to load the Theme (that's what I suggest in that blog post) we could include the loading in some Mezzanine middleware.

Thoughts?

AlexHill:
Like Josh, I was going to say that this is only really worthwhile IMO if integrated into the media library so that users can upload the logo directly – giving them a text field doesn't help beginners out that much if they also have to work out how to upload the file, and what path to enter.

Josh's solution sounds much better than that. A built-in generalised way to expose editable, translatable snippets of "theme" content like this would be great and would let us remove non-setting settings like site title and tagline, and remove support for translatable settings too.

It might be good to store the theme snippets in a key-value style as editable settings are, so users could change between themes without any schema/migration problems.

orotau:
Hi Josh
I have just read your blog post a couple of times and I found it easy to 
follow.
It seems to 'fit' well with Mezzanine, so I would support your 
suggestion as far as my current knowledge allows :-)
Thanks
g

On 20/05/15 12:28, Josh Cartmell wrote:
>
> I really like the idea of making the logo user editable, but I think 
> storing the path to it in the settings is a bit brittle. I think it's 
> easy for a developer/computer savvy person, but I think finding the 
> path to an image may go over the heads of a lot of admin users.
>
> My preferred method for making content (like a logo) backend editable 
> is this, 
> http://bitofpixels.com/blog/on-singletonadmins-and-sitewidecontent-editing-sitewide-content-in-mezzanines-admin/
>
> Basically a model that is a subclass of SiteRelated (let's call it 
> Theme) and is registered with the admin using the SingletonAdmin 
> class. The benefits I see are:
>
>  1. Familiar format to admin users
>  2. Familiar widgets for admin users (i.e. the logo can use a
>     FileField and have the nice image selector widget)
>  3. The editable template tag works (afaik you can't use editable with
>     settings)
>  4. Like Ken said, the settings can be overwhelming to the
>     uninitiated. I find it helpful to keep logic in the settings, and
>     content in my Theme model. There is probably some overlap, but I
>     think it's helpful to keep them as separate as possible
>
> I think something like what I outline in that blog post could be worth 
> including in Mezzanine. We could have an AbstractTheme model that just 
> included the logo (maybe title and tagline) and then a default Theme 
> that just inherited from that. Developers could override the theme, 
> inheriting from the base model if they wanted. Instead of using a 
> templatetag to load the Theme (that's what I suggest in that blog 
> post) we could include the loading in some Mezzanine middleware.
>
> Thoughts?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/stephenmcd/mezzanine/pull/1298#issuecomment-103705465>.
>

-- 
T : 021 081 71732

Kniyl:
I really like the singleton approach too.

One quick thinking though: it might be difficult to have both key/values pairs stored in the database in a settings-like fashion and an easy way to integrate the editable template_tag with.

joshcartme:
I'd be in favor of using regular models, not key-value pairs.  The biggest reason is that models are familiar to Django developers so it would be much easier to customize that way; i.e. custom admin classes, custom admin form classes, model validation, etc...

Theme may not have been the perfect word (at least how I use it).  As an example, here's one of these models from a project I worked on:

    class SiteConfiguration(SiteRelated):
        '''
        A model to edit sitewide content
        '''
        logo = FileField(verbose_name=_("Logo"),
            upload_to=upload_to("theme.SiteConfiguration.logo", "theme"),
            format="Image", max_length=255)
        linkedin_link = models.CharField(max_length=2000,
            help_text="If present a Linked in icon linking here will show "
                      "in the header", blank=True)
        facebook_link = models.CharField(max_length=2000,
            help_text="If present a Facebook icon linking here will show "
                      "in the header", blank=True)
        google_plus_link = models.CharField(max_length=2000,
            help_text="If present a Google Plus icon linking here will show "
                      "in the header", blank=True)
        twitter_link = models.CharField(max_length=2000,
            help_text="If present a Twitter icon linking here will show "
                      "in the header", blank=True)
        class Meta:
            verbose_name = _('Site Configuration')
            verbose_name_plural = _('Site Configuration')

I mostly use it for content in the header and footer of sites that is present on every page, but it's also desirable for it to be admin editable.

I think the question to answer is: Does this belong in Mezzanine?  It's really easy to do outside of Mezzanine.  My vote would be yes because it would provide a bit more configurability out of the box and it would provide a good example of the usefulness of the SingletonAdmin.

stephenmcd:
Benefit of kv pairs (eg mezzanine.conf) is that no migrations need to occur. My feeling is that the fields on that example `SiteConfiguration` model would vary widely per site.

AlexHill:
Depends what we mean by "theme" – if we want to pursue a goal of users being able to switch themes in the admin, migrations are a deal-breaker. But if we mean a theme app that the site developer installs once at the start of a project or develops themselves, requiring migrations is more OK. Sounds like that's what you have in mind Josh.

Using normal models would also mean we could rely on standard modeltranslation machinery and remove the translatable settings code. I can't currently think of a setting that a) requires translation and b) doesn't make more sense as a theme-related setting.

joshcartme:
@stephenmcd I agree it is very project specific, that was just meant as an example of the way I use these Theme/SiteConfiguration models.  I think because it is so specific to each project creating migrations isn't onerous.

@AlexHill I think you make a good point about trying to distinguish what the goals are.  I always use it inline with your second example.  The more I think about it, at least for my use case, theme is a bit of a misnomer, I just use it as editable sitewide content.

Kniyl:
@AlexHill just in case, I made the SHOP_CURRENCY_LOCALE a translatable setting. And I don't think it makes more sense as theme-related setting. But if we decide to go on the way of a model that controls some of the content in the templates, and given that it simplifies things, it might go in there too.

Once again, I really like this approach and if we decide to integrate it more tightly into mezzanine I can put some time into it. It could show the path of easy and flexible theming to mezzanine users.

Anyway, I think that this discussion goes in the direction of “we won't merge this PR, at least not as is” so why don't we close it and inaugurate the mezzanine-developers list to see what all this can create?

b3h74d:
After reading these comments, I wonder how such a simple and usefull change, which follows base Mezzanine approch of settings to implement a logo, makes this long discussion. 

It is simple as other settings, it helps a lot of users I think, and ...If you do not like it, please show  a better approach to do this, or we can accept this until a better approach comes to mind, maybe?

stephenmcd:
Hi @b3h74d - I know it's a bit frustrating! Think of it as quality assurance - if such a small change is scrutinized so heavily, you can have some confidence in the overall quality of the project :-)

Personally I agree with the comments above - having a filesystem path edited via a text box in the settings admin isn't really acceptable. It would need to be a file field, but then the big question is indeed opened up (as discussed), as to whether a file represents a setting, or some other alternative should come into play for managing theme based things.

Please be patient, this is a healthy process.

b3h74d:
Hi @stephenmcd , It's totaly ok to me, I know that community would find the best way as always, at the end most of us benefit from implemention of this, the best way.

Kniyl:
Same here. I’m sorry if my comment seemed a bit harsh.

The point was not to denigrate your work, but rather acknowledge the fact that it opens interesting questions that might be better adressed and integrated by other means.

And as we are slowly building up as a team, I was trying to get things started together.

b3h74d:
As a point, it is actually a url, but relative address added by default, it is better to implement it as absolute url by default and if it is simple, a button next to field to choose a file from media files...

kenbolton:
Hi @b3h74d! As you can see from this long discussion, one developer's useful feature is another developer's anti-pattern. Your contribution has sparked a discussion that will have a positive impact on the future of Mezzanine, though, and that is more valuable than lines of code merged.

One suggestion, @b3h74d, is to publish your implementation as a third-party Mezzanine app. Then you don't have to hear anything from us. If you do, please be sure to modify the README for Mezzanine to list your app and do a pull request.

rsadwick:
Looks good but I agree that it should be a file field.  All of my Mezzanine powered sites use AWS s3 for serving content - manually uploading and throwing in a relative path isn't a good solution for the sites I'm running since I'd manually have to set the caching.

I don't mind this setting but @kenbolton brings up some good points about the amount that is already there by default.

PR #1305, opened 2015-05-22
Make the settings context processor lazy.

frankier:
This way a database hit is avoided on pages where the settings aren't used. This is particularly useful when Mezzanine is being used as part of a larger Django site where there isn't the need for the settings in every template.

### DIFF
--- a/mezzanine/conf/context_processors.py
+++ b/mezzanine/conf/context_processors.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from django.utils.functional import lazy
 from mezzanine.utils.cache import (cache_key_prefix, cache_installed,
                                    cache_get, cache_set)
 
@@ -59,4 +60,4 @@ def settings(request=None):
             settings_dict["MEZZANINE_ADMIN_PREFIX"] = "admin/"
         lazy_settings.settings_dict = settings_dict
         return settings_dict
-    return {"settings": lazy_settings}
+    return {"settings": lazy(lazy_settings, TemplateSettings)()}
--- a/mezzanine/conf/context_processors.py
+++ b/mezzanine/conf/context_processors.py
@@ -29,32 +29,34 @@ class TemplateSettings(dict):
 
 
 def settings(request=None):
-    """
-    Add the settings object to the template context.
-    """
-    from mezzanine.conf import settings
-    settings_dict = None
-    cache_settings = request and cache_installed()
-    if cache_settings:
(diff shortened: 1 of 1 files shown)


Kniyl:
According to the Travis build, your change prevent accessing context['settings'] as the dictionnary it should be.

Why not define a lazy_settings on top of settings, and having to explicitly choose which one to use in the settings.py file?
Something along the lines of:
```python
def lazy_settings(request=None):
  def settings_getter():
    if not hasattr(lazy_settings, 'settings_dict'):
      lazy_settings.settings_dict = settings['settings']
    return lazy_settings.settings_dict
  return {'settings': settings_getter}
```

frankier:
I found a helper in Django to create a lazy proxy object. I thought it was fairly idiomatic to put nullary functions in the template context, but obviously this doesn't work so well when templatetags get involved and we go back into python land.

Anyway, it's got the same interface as before now.

AlexHill:
Hey Frankie, thanks for the patch. I think this should be unnecessary – Mezzanine's settings object is lazy itself in that settings are only loaded from the database the first time an editable setting is accessed.

Have a look here: https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/conf/__init__.py#L191

You've submitted this patch so obviously you've noticed an extra unwanted DB call – can you maybe put a breakpoint at that line and see what's triggering it?

frankier:
The settings object is lazy, but the settings query will be evaluated if any of the TEMPLATE_ACCESSIBLE_SETTINGS are editable. In my case some of these are needed some of the time (eg for admins, editiable_loader.html) so they should be accessible in the templates - yet in the normal case they're not needed. This way the evaluation of the settings/the database query is deferred even further - letting the template control what happens.

Perhaps there's a more elegant way of doing this?

AlexHill:
Hey @frankier, sorry for the delay.

> but the settings query will be evaluated if any of the TEMPLATE_ACCESSIBLE_SETTINGS are editable

Gotcha, missed that.

> Perhaps there's a more elegant way of doing this?

I think so:  ``TemplateSettings`` could wrap ``mezzanine.conf.settings``, and encapsulate this logic in its ``__getattr__`` method. That way we don't access the settings until they're needed, and the laziness still lives only in ``Settings``.

While we're at it, we could probably just get rid of the context settings caching entirely if we adopted that approach.

AlexHill:
@frankier something like this: https://github.com/stephenmcd/mezzanine/compare/master...AlexHill:template_settings

frankier:
Looks good to me.

PR #1316, opened 2015-05-26
Make mezzanine.utils.urls.path_to_slug pluggable.

frankier:
This is needed for sites with a custom i18n url scheme (and might be useful for other types of customisation too).


Kniyl:
I’d rather write it as:
```python
def path_to_slug(path):
    _path_to_slug = settings.PATH_TO_SLUG
    if _path_to_slug:
        return import_dotted_path(_path_to_slug)(path)
    # regular path_to_slug_definition
```
to avoid the cost of an extra import for each page if it’s not needed. (Given that ```PATH_TO_SLUG``` defaults to the empty string.)

But it might also be unnecessary, not sure of this one.

frankier:
It's using the same pattern as SLUGIFY. This makes it easy to implement your own path_to_slug in terms of mezzanine_path_to_slug.

import_dotted_path uses importlib's import_module. This will fetch the module from Python's internal module cache after the first import, exactly the same as Python's import statement.

stephenmcd:
Can you talk more about the use case? Not even clear why this is needed.

frankier:
The current path_to_slug strips a language code, like en or en-us or en-gb from the url. It's possible for a site (like the one I'm working on) to have a different url scheme, eg using ccTLD codes in urls like /uk/ /fi, /de// and so on. It's probably impossible to cover every possible case in Mezzanine so making this an explicit extension point seems nicer.

Currently we're monkeypatching this function but I'm trying to reduce the number of monkeypatches in our code base. I've seen many different types of i18n urls around the place so I think other people could well have the same use case.

stephenmcd:
Just reluctant to add yet another setting for what seems like an edge-case.

frankier:
Sure we'll just carry on monkeypatching it then.

PR #1318, opened 2015-05-27
Add index on publish_date

frankier:
I've tested this and it seems to half query runtime eg, for retrieving the latest blog post. Explain with Postgres confirms the index is being used.


stephenmcd:
Great idea - thanks Frankie

Kniyl:
We might as well provide a migration file for ```cartridge.shop```

stephenmcd:
Done in https://github.com/stephenmcd/cartridge/commit/786afc415dc2a194ef1a8ec24e4d632ea477699a

PR #1323, opened 2015-05-30
Load editable settings once per request

dsanders11:
**NOTE:** This is not a full pull request, I don't expect this to be merged. I just want to start a discussion, and the best way to start a discussion is bring working code to the table.

Currently editable settings are loaded from the database any time `settings.use_editable()` is called. This can be several times during the same request, depending on the view and functions called. Some views I see it happen 4 or 5 times.

It seems undesirable to load editable settings from the database several times per request for several reasons:
 1. Not idempotent - can lead to weird results for a request that is being processed when an editable setting is changed
 2. Inefficient - extra trips to the database that can be eliminated help with performance
 3. Inconsistent - settings available in template contexts are currently cached in the cache long-term while the settings object may reload settings from the database, causing a mismatch between the two

I suggest that editable settings are at the very least cached per-request, as this pull request does (in an ugly way). I think a more complete solution would be to also put the editable settings in the cache like the context settings are, and invalidate the key if modified from the admin area (as #1212 implemented).

The combination of a per-request and global caching would resolve the three issues I mentioned above. The per-request caching would first check the cache, then hit the DB and populate the cache if necessary. This fulfills 1 as a single request will be entirely rendered with one view of the settings, and 2 as database dips will only be necessary upon the double caching miss.

It mostly resolves 3, although the separate caching may still lead to an inconsistency between the template context and the request's view of the settings if the settings are changed from the admin area mid-request (resulting in the template context having the newer settings and the request having the older settings).

Perhaps a most complete solution would have the settings context processor also check the request's cache of the editable settings and use those if available.

Thoughts?

### DIFF
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -16,6 +16,7 @@ from django.utils.functional import Promise
 from django.utils.module_loading import module_has_submodule
 
 from mezzanine import __version__
+from mezzanine.core.request import current_request
 
 
 registry = {}
@@ -115,6 +116,12 @@ class Settings(object):
         """
         self._loaded = True
         self._editable_cache = {}
+        
+        request = current_request()
+        
+        if request:
+            request._loaded_editable_settings = False
+            request._editable_settings = {}
 
     def use_editable(self):
         """
@@ -177,8 +184,15 @@ class Settings(object):
                  % ", ".join(conflicting_settings))
         self._editable_cache = new_cache
         self._loaded = True
+        
+        request = current_request()
+        
+        if request:
+            request._loaded_editable_settings = True
+            request._editable_settings = new_cache
 
     def __getattr__(self, name):
+        request = current_request()
 
         # Lookup name as a registered setting or a Django setting.
         try:
@@ -186,8 +200,13 @@ class Settings(object):
         except KeyError:
             return getattr(django_settings, name)
 
+        loaded = self._loaded
+        
+        if request:
+            loaded = getattr(request, "_loaded_editable_settings", False)
+
         # First access for an editable setting - load from DB into cache.
-        if setting["editable"] and not self._loaded:
+        if setting["editable"] and not loaded:
             self._load()
 
         # Use cached editable setting if found, otherwise use the
@@ -195,7 +214,10 @@ class Settings(object):
         # exists, finally falling back to the default defined when
         # registered.
         try:
-            return self._editable_cache[name]
+            if request and hasattr(request, "_editable_settings"):
+                return request._editable_settings[name]
+            else:
+                return self._editable_cache[name]
         except KeyError:
             return getattr(django_settings, name, setting["default"])
 
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -116,9 +116,9 @@ class Settings(object):
         """
         self._loaded = True
         self._editable_cache = {}
-        
+
         request = current_request()
-        
+
         if request:
             request._loaded_editable_settings = False
             request._editable_settings = {}
@@ -184,9 +184,9 @@ class Settings(object):
                  % ", ".join(conflicting_settings))
         self._editable_cache = new_cache
         self._loaded = True
-        
+
         request = current_request()
-        
+
         if request:
             request._loaded_editable_settings = True
             request._editable_settings = new_cache
@@ -201,7 +201,7 @@ class Settings(object):
             return getattr(django_settings, name)
 
         loaded = self._loaded
-        
+
         if request:
             loaded = getattr(request, "_loaded_editable_settings", False)
 


stephenmcd:
I like it - always wanted to get rid of `use_editable` and just have it Work. 

Keen to hear others' thoughts.

dsanders11:
"Working code" - I did the edits here on GitHub then tested, but apparently GitHub likes whitespace on blank lines, heh.

@stephenmcd, I always assumed the explicit call to `use_editable` was required to work around some issue. I'd also like to kill it if possible, but wasn't sure of the context of it's origin so I didn't go so far as to propose that.

AlexHill:
*Not* calling ``use_editable()`` guarantees you won't hit the database. That's potentially useful but I doubt is actually being relied upon in practise.

Do you think it would be worth adding a setting to switch editable settings off altogether? I suppose that would include removing them from the admin.

AlexHill:
To be clear, I think getting rid of ``use_editable()`` makes absolute sense and I don't think there needs to be a setting. If anybody really wants to stop ``Settings`` from hitting the database, they can just set the ``_loaded`` flag manually.

dsanders11:
@AlexHill, yea, and between the request caching and cache key, best case it hits the DB every however long we set the cache key for, worst case it hits it once per request. I don't think anyone will get too upset over one DB dip.

The turn off for editable settings altogether is a good idea. I have some ideas on implementing this all which I'll try to hammer out.

AlexHill:
That said there are a couple of things I don't like so much about the implementation:
 1. ``current_request()`` – but if we want to be smarter about DB accesses, which is kind of the whole point of this change, we probably need it.
 2. setting even more attributes on the request object. We should constrain this state to the ``Settings`` class – basically, let's keep all this magic in one place. We can do that with a ``WeakKeyDictionary`` mapping requests to settings dicts.

AlexHill:
See what you think: https://github.com/stephenmcd/mezzanine/compare/master...AlexHill:smarter_settings

dsanders11:
@AlexHill, interesting start, I'll read it over more tomorrow when my brain is less frazzled.

~~One thing I'll note from the preliminary skim, though, is that I think `def _editable(self, request=None)` should be part of the public interface, so that the settings context processor can call it to layer in the cached settings for the request if there are any cached on it. That goes to point 3 in my original write-up.~~

I think actually the whole caching implementation for `mezzanine.conf.context_processor.settings` should be moved to `class Settings` and the context processor will simplify to grabbing the available template settings from `class Settings` which will encapsulate the caching behavior.

Also, the last failure on my bogus code earlier was because the thread local from `current_request()` lived throughout the test run causing cached settings. I understand your aversion to it, it's icky, but it's already in the code base, so may as well use it. We may want to use the [request_finished](https://docs.djangoproject.com/en/1.7/ref/signals/#django.core.signals.request_finished) signal to try clearing out any stale values, or look into making it a weak ref as well.

Final thought for tonight, it also occurs to me that some people may be using the model ability of settings to modify editable settings programmatically. Might be a good idea to listen to the save signal for the Settings model and invalidate the caches when that occurs. Might allow us to ditch the special case of the admin code knowing the cache key name so that it can reach over and invalidate the cached settings.

AlexHill:
> I think actually the whole caching implementation for `mezzanine.conf.context_processor.settings` should be moved to `class Settings`

I agree, but I think the context settings caching could just be removed once this is all in place. I actually just proposed that today here: https://github.com/stephenmcd/mezzanine/pull/1305#issuecomment-107113410

> I understand your aversion to it, it's icky, but it's already in the code base, so may as well use it.

Yep, and there's just no alternative if we want to fix the problems you outlined in your initial post and maintain the normal ``settings.SETTING`` interface.

> We may want to use the [request_finished](https://docs.djangoproject.com/en/1.7/ref/signals/#django.core.signals.request_finished) signal to try clearing out any stale values

The middleware that stores the current request as a thread-local has [some commented-out code](https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/request.py#L24) that would do this at the end of each request – @stephenmcd any reason we can't just re-enable that?

> Might be a good idea to listen to the save signal for the Settings model and invalidate the caches when that occurs.

I wanted to do this the last time I made some changes to the editable settings code but as Steve pointed out it won't work if you're running in more than one process: https://github.com/stephenmcd/mezzanine/pull/1126#issuecomment-58299185

> Might allow us to ditch the special case of the admin code knowing the cache key name so that it can reach over and invalidate the cached settings.

Could you point me to that code please?

dsanders11:
> The middleware that stores the current request as a thread-local has some commented-out code that would do this at the end of each request – @stephenmcd any reason we can't just re-enable that?

I saw that too, I imagine it might have something to do with the middleware order?

> I wanted to do this the last time I made some changes to the editable settings code but as Steve pointed out it won't work if you're running in more than one process: #1126 (comment)

That's the nice thing about the cached version being on the request and in the shared cache. Invalidating the shared cache key invalidates it for all processes, and we don't care about the per-request version because we want it to remain and requests are short lived.

> Could you point me to that code please?

#1212

AlexHill:
> Invalidating the shared cache key invalidates it for all processes, and we don't care about the per-request version because we want it to remain and requests are short lived.

Sorry yes, that's true. There is still the out-of-request case, but that should only happen on startup (and in tests as we have both seen, but we can deal with that).

I'm not sure that we'll need the context settings to be externally cached at all once this and #1305 are resolved. What reason is there to cache specifically these settings and not others?

dsanders11:
> I'm not sure that we'll need the context settings to be externally cached at all once this and #1305 are resolved. What reason is there to cache specifically these settings and not others?

Agreed, all of the external caching would be done in `class Settings`, helping to keep a nice consistent view of settings during a single request. With the per-request cache, and context settings using `class Setings`, it should be impossible to have a request see two different values for a setting (like it currently can) unless the settings were explicitly cleared. I'd suggest any default clear cache behavior would clear the external cache and leave the per-request cache unless explicitly asked for.

AlexHill:
I'm proposing dropping external (i.e. shared) caching for settings entirely – agree with everything else.

dsanders11:
@AlexHill, I'd really prefer to have the external caching of settings, at least as an option (it could be off by default).

Out of the box Mezzanine uses quite a bit of DB queries (especially due to Settings as discussed here) if a user is logged in, even on simple GET requests. Being able to tweak settings to make a no DB hit experience when simply browsing the site is a huge performance benefit and realistically allows the web server to sit further from the DB with less of a performance burden.

Most aspects of Mezzanine that hit the DB I can cache without needing to muck with internals very much (can put a `{% cache %}{% endcache %}` block in templates), but Settings is deep enough into view code that you can't easily prevent the DB hit. Since Settings change so rarely (and in a controlled manner, via the admin) it makes sense to cache them. It's also not something I can easily change in my project without some monkey patching due to how settings are imported (or a full Mezzanine fork, which I try to avoid for maintenance reasons).

We could allow optional caching through something like:

````python
# mezzanine/conf/__init__.py
from django.core.cache import caches

try:
    settings_cache = caches["mezzanine_settings"]
except InvalidCacheBackendError:
    settings_cache = None

# settings.py

CACHES = {
    "default": ....
}

# Enable Mezzanine settings caching
CACHES["mezzanine_settings"] = CACHES["default"]
````

AlexHill:
I'm not opposed to that idea, but that would be a new feature – I just want to get rid of the special-case caching for context settings for now.

Once this and #1305 are resolved, the context settings cache would only save a DB call if no editable settings are accessed in Python code in the entire request cycle, which is not impossible but certainly fairly rare.

stephenmcd:
Apologies for not having red this thread in detail, nor the code, but I've had an insanely busy week.

Alex you briefly mentioned not liking reliance on the `current_request` - I don't think getting rid of `use_editable` would be possible without it. Is there another way? BTW I think that bridge has been crossed already, with Cartridge's cart and all the multi-site support in Mezzanine. 

So if we're comfortable with that, I think a ton of the code currently in the settings implementation can go - everything around having a cache that can be flushed and reloaded, namely `use_editable` etc.Which means this PR is probably far from complete. I think it might also make the other tangental PR #1325 somewhat redundant. 

Just my quick thoughts for now - again I haven't gone over this thoroughly so apologies if I've missed something obvious.

AlexHill:
These changes are now a part of #1326, let's move the discussion there.

PR #1324, opened 2015-05-31
New management command to nicely dump settings

dsanders11:
I coded up a management command to dump out all the registered settings and nicely format them with their description to file. Should be handy for keeping up to date with new settings and make it easy to pump out new templates.

The settings are grouped under the starting package, so it'll group Mezzanine settings all under one section. Also works great for third-party apps if they use `register_setting`.

Output is a fully functional settings module, but with everything commented out (you can uncomment the values by changing PREFIX in the file, an improvement would be to make that a command line arg). It passes PEP8 and everything.

The formatting is mostly what humans would do, it gets a little weird with tuples and lists but I did the best I could (was a pain in the ass) without going overboard. They're not ideally formatted, but they're good enough, and not that common.

Give 'er a whirl.

### DIFF
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -11,6 +11,8 @@ from functools import partial
 from importlib import import_module
 from warnings import warn
 
+import inspect
+
 from django.conf import settings as django_settings
 from django.utils.functional import Promise
 from django.utils.module_loading import module_has_submodule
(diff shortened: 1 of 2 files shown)


stephenmcd:
Looks neat - might serve better as a third party app though.

dsanders11:
Which would be dandy - except for the small change it requires to `register_setting` to identify which app registered the setting. :-)

stephenmcd:
I'm honestly not keen on that small change either - introspection hacks like that tend to be very brittle, although I can't qualify that any further.

I'd suggest making this a separate app, without the "app" variable per setting. Then we can announce it on the mailing list and add it to the list of apps.

BTW the reason for not including it is the amount of code it would add that needs to be maintained, for a feature that while certainly handy, probably wouldn't be used by most people.

PR #1325, opened 2015-06-03
Make setting context processor lazy

frankier:
Previously https://github.com/stephenmcd/mezzanine/pull/1305


stephenmcd:
This may become redundant based on #1323 

I think both PRs are trying to achieve something similar, though both might not necessarily be correct - I personally haven't had the time available to dive in properly yet.

Just pointing this out so both threads of work can cooperate.

AlexHill:
Yep, it will make sense to make both these changes at once. I have my work for both of these in one branch locally, I'll push it in a sec.

(BTW I had no idea you could open a PR from someone else's branch...nifty.)

AlexHill:
Incorporated into #1326, let's move the discussion there.

PR #1326, opened 2015-06-05
Lazier, more consistent editable settings

AlexHill:
This incorporates the changes in behaviour proposed by @dsanders11 in #1323 and @frankier in #1305. Can you two please test this branch in your projects to confirm that it does what it's meant to?

Overview:
 * we now automatically use database-backed settings.
  * ``settings.use_editable()`` is out (it's still there, but comes with a deprecation message).
  * ``settings.clear_cache()`` is in, for when code needs to explicitly refresh database-backed settings
 * settings are consistent for the duration of each request (unless ``settings.clear_cache()`` is called)
 * the template context settings cache is gone, and the associated changes from #1212 reverted
 * a few more tests now nail down the behaviour of editable settings more precisely
 * settings accessed in templates are now lazily fetched, instead of being fetched upfront

@stephenmcd would be great if you could cast an eye over this when you have some time.

### DIFF
--- a/mezzanine/accounts/forms.py
+++ b/mezzanine/accounts/forms.py
@@ -221,7 +221,6 @@ class ProfileForm(Html5Mixin, forms.ModelForm):
             pass
 
         if self._signup:
-            settings.use_editable()
             if (settings.ACCOUNTS_VERIFICATION_REQUIRED or
                     settings.ACCOUNTS_APPROVAL_REQUIRED):
                 user.is_active = False
--- a/mezzanine/blog/feeds.py
+++ b/mezzanine/blog/feeds.py
@@ -45,7 +45,6 @@ class PostsRSS(Feed):
         else:
             self._public = not page.login_required
         if self._public:
-            settings.use_editable()
             if page is not None:
                 self._title = "%s | %s" % (page.title, settings.SITE_TITLE)
                 self._description = strip_tags(page.description)
--- a/mezzanine/blog/views.py
+++ b/mezzanine/blog/views.py
@@ -25,7 +25,6 @@ def blog_post_list(request, tag=None, year=None, month=None, username=None,
     ``blog/blog_post_list_XXX.html`` where ``XXX`` is either the
     category slug or author's username if given.
     """
-    settings.use_editable()
     templates = []
     blog_posts = BlogPost.objects.published(for_user=request.user)
     if tag is not None:
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -94,11 +94,11 @@ class Settings(object):
     of access for all settings.
     """
 
-    class Sentinel(object):
-        """A Weakly-referable wrapper of 'object'."""
+    class Placeholder(object):
+        """A Weakly-referable wrapper of ``object``."""
         pass
 
-    NULL_REQUEST = Sentinel()
+    NULL_REQUEST = Placeholder()
 
     # These functions map setting types to the functions that should be
     # used to convert them from the Unicode string stored in the database.
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -232,10 +232,9 @@ class Settings(object):
         if not setting["editable"]:
             return getattr(django_settings, name, setting["default"])
 
-        # Use cached editable setting if found, otherwise use the
-        # value defined in the project's settings.py module if it
-        # exists, finally falling back to the default defined when
-        # registered.
+        # Use cached editable setting if found, otherwise use the value
+        # defined in the project's settings.py module if it exists,
+        # finally falling back to the default defined when registered.
         editable_cache = self._get_editable(request=self._current_request)
         try:
             return editable_cache[name]
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -5,6 +5,7 @@ or Django itself. Settings can also be made editable via the admin.
 """
 
 from __future__ import unicode_literals
+from weakref import WeakKeyDictionary
 from future.builtins import bytes, str
 
 from functools import partial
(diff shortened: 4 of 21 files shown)


AlexHill:
@dsanders11, quoting myself from https://github.com/stephenmcd/mezzanine/pull/1323#issuecomment-107280876 where I wrote the following:

> [After this change], the context settings cache would only save a DB call if no editable settings were accessed in Python code in the entire request cycle, which is not impossible but certainly fairly rare.

Do you agree with that analysis? I think an external settings cache is not a bad idea, but I don't think we're losing anything by removing the template settings cache here.

stephenmcd:
Few thoughts:

* This is minor, but do we really need `__setattr__` and `__delattr__` other than for completeness? I don't think we set or delete settings externally anywhere, nor should they be encouraged as an API. My preference would be to err on the side of less code.

* I don't understand the motivation with storing multiple caches mapped to request objects, in fact it feels like the opposite of what I had in mind and had vaguely gleaned from the original PR - which is basically looking at the current request object, checking for a settings attribute, and assigning the new (db refreshed / lazy) cache to it. It seems like that approach might achieve the same thing, but with a more lightweight approach. That's a similar model to what we've done with Cartridge carts and multi-tenant site IDs - check for an attr on the request and assign it if missing. 

* This is meta, but we should probably squash the commits before merging, while we're still relying on an auto-generated changelog.

AlexHill:
Thanks Steve.

> do we really need __setattr__ and __delattr__ other than for completeness?

Not really, but I find the whole system is easier to think about if ``mezzanine.conf.settings`` only ever deals with editable settings, and ``django.conf.settings`` always handles normal attribute settings. Setting attributes on ``mezzanine.conf.settings`` also suppresses the settings conflict warning in some cases.

> I don't understand the motivation with storing multiple caches mapped to request objects, in fact it feels like the opposite of what I had in mind and had vaguely gleaned from the original PR

I'd describe it as inverted, rather than opposite: either way, a new dict is created for each request, but in this code it's stored in ``settings._editable_caches`` rather than ``request.__dict__``. It's O(1) either way.

``request.cart`` is used all over the place throughout Cartridge and user code, and has an analogy to ``request.user``, so I think it belongs there. These cached settings dicts are just an implementation detail of ``Settings`` which I think shouldn't leak out if it can be prevented.

> This is meta, but we should probably squash the commits before merging, while we're still relying on an auto-generated changelog.

Definitely.

stephenmcd:
> I'd describe it as inverted, rather than opposite: either way, a new dict is created for each request, but in this code it's stored in settings._editable_caches rather than request.__dict__. It's O(1) either way.

Inverted is much more correct yeah :-) And thinking about how there isn't a corresponding `request.settings` makes sense in describing the motivation. 

I'm still a bit unsure about the approach though. We've gone from having one strategy used for global request variables (assigning an attribute to the request object, as per cart, site_id), to two (that and using `WeakKeyDictionary`) - would it be simpler to just stick with a single approach? It's also a trodden path that we know works, while `WeakKeyDictionary` is an unknown. That's not to say it wouldn't work though.

AlexHill:
I guess it's just a difference in perspective. For me, it's more important that ``Settings`` is self-contained. ``request`` can spend a lot of time bouncing around in code we didn't write, so as a matter of principle, I think the less we touch it, the better. ``request.cart`` is more or less a public interface, so in my head that falls in a different category of thing. (I'm not familiar enough with the site code to comment on that.)

> It's also a trodden path that we know works, while ``WeakKeyDictionary`` is an unknown.

Hear that loud and clear.

For what it's worth, I've used ``WeakKeyDictionary`` to associate data with requests before with success. I just gave my test server a thrashing and kept track of the size of settings.editable_caches and it never got above 20, so it's not as though heaps of these are hanging around.

I think I've said enough – happy to change it if your position is unchanged :)

stephenmcd:
If you've used it before and tested it then let's go with it - I guess I'm just a little thrown off because I've never seen it before. It's an interesting pattern - glad to have learnt about it :-)

Thanks Alex.

AlexHill:
Okay, cool.

Something else has come up. This code (in master) is from ``generic/admin.py``:

```python
generic_comments = getattr(settings, "COMMENTS_APP", "") == "mezzanine.generic"
if generic_comments and not settings.COMMENTS_DISQUS_SHORTNAME:
    admin.site.register(ThreadedComment, ThreadedCommentAdmin)
```

Couple of problems with this:
 * ``COMMENTS_DISQUS_SHORTNAME`` is an editable setting, so with editable-by-default this fails when you run ``migrate`` initially
 * In uwsgi/nginx at least, having an editable setting in module-level code makes Django open a connection to the database before it forks, which causes problems

I've hacked around the first one in this branch but the second is harder.

Possible solutions:
 1. Outside of requests (i.e. at module import time), maintain the current ``use_editable()`` behaviour, so the database isn't used unless specifically requested
 2. In this specific case, use ``django.conf.settings`` instead to circumvent the database
 3. Something else?

frankier:
Seems like good work. Could something else be an AppConfig since Django <1.7 doesn't need to be supported? It works for me in that it prevents an SQL query on pages that don't use any editable settings, which was my original problem. :shipit:

AlexHill:
I think that will have the same problem, because ``django.apps.populate()`` is called in ``django.setup()``, which is called in ``wsgi.py``.

To summarise the problem: database handles can't survive forks. Some host applications, like uwsgi, fork after application startup by default. Therefore fetching editable settings from the database during startup can cause database exceptions.

Currently, this value isn't pulled from the database anyway, because there's no call to ``use_editable()``. I say we just use ``django.conf.settings`` for this logic. It's very simple and existing behaviour won't change.

stephenmcd:
Yep agreed - from my perspective, let's try and keep the iterations small, and this current PR is a pretty big leap as is.

stephenmcd:
BTW I'm just about to start testing this.

stephenmcd:
I've tested this:

- Site runs fine
- DB queries are limited to one per request
- Updating DB settings take affect

Looks good!

AlexHill:
Last commit makes the changes discussed above.

PR #1328, opened 2015-06-11
Update jquery to 1.7.2, jquery-mobile to 1.2.1.

nikolas:
In an effort to gradually bring mezzanine's javascript up to date,
this commit updates mezzanine's default jquery version to the
latest in the 1.7.x series (which is 1.7.2).

Because the 1.7.2 release notes state that this version
should be used with jQuery Mobile >= version 1.1, I've updated
jQuery Mobile as well.
http://blog.jquery.com/2012/03/21/jquery-1-7-2-released/


stephenmcd:
Thanks Nikolas

PR #1330, opened 2015-06-12
update bootstrap to 3.3.4

gradel:


jerivas:
Hey @gradel. The problem with this is that the latest version of Bootstrap requires at least jQuery 1.9.1. Mezzanine ships with jQuery 1.7.x due to some plugins it uses in the admin. For this to work, you'd need to upgrade jQuery to a newer version (probably should go all the way to the latest version in the 1.x branch while we're at it) and then test the admin to make sure nothing breaks. You'll also need to test the inline editor and more.

jerivas:
Also, here's some discussion for reference: https://groups.google.com/forum/#!searchin/mezzanine-users/upgrade$20jquery/mezzanine-users/jlcdcReUwdM/YjQ6pJPj3z8J

gradel:
Ups, sry, overlooking a lot again ;)

PR #1337, opened 2015-06-22
Allow overriding fabric ENVS, create sync task.

joshcartme:
This is a bit of a work in progress.

Update the FABRIC dictionary to be expected to look like this (the old style will still work as well):

```
FABRIC = {
    "DEFAULT": {
        "SSH_USER": "", # SSH username for host deploying to
        "HOSTS": ALLOWED_HOSTS[:1], # List of hosts to deploy to (eg, first host)
        "DOMAINS": ALLOWED_HOSTS, # Domains for public site
        "REPO_URL": "git@github.joshcartme/amezzproj.git", # Project's repo URL
        "VIRTUALENV_HOME":  "", # Absolute remote path for virtualenvs
        "PROJECT_NAME": "", # Unique identifier for project
        "REQUIREMENTS_PATH": "requirements.txt", # Project's pip requirements
        "LOCALE": "en_US.UTF-8", # Should end with ".UTF-8"
        "DB_PASS": "this-is-not-a-real-password", # Live database password
        "ADMIN_PASS": "this-is-not-a-real-password", # Live admin user password
        "SECRET_KEY": SECRET_KEY,
        "NEVERCACHE_KEY": NEVERCACHE_KEY,
    },
    'DEV': {
        "SSH_USER": "dev",
        "HOSTS": ['dev.example.com'],
        "VIRTUALENV_HOME": "/home/dev",
    },
    'PROD': {
        "SSH_USER": "prod",
        "HOSTS": ['www.example.com'],
        "VIRTUALENV_HOME": "/home/prod",
        "DB_PASS": "xyz---abc---zyx",
    }
}
```

In the above `DEV` and `PROD` are optional dictionaries that can override values in the `DEFAULT` dictionary.  `DEV` and `PROD` are not required and any other names or number of dictionaries can be used.

To have a non default dictionary override pass `--set FABENV=ENV` to any fabric command where ENV is a key in the `FABRIC` dictionary. **For example**:

```
$ fab all --set FABENV=DEV
```

would call the `all` task and override any values in `DEFAULT` that are also in `DEV` with the values from `DEV`.

If no overrides are desired call fabric commands like normal, i.e.:

```
$ fab all
```

My motivation for the above changes was to create a sync command that will sync the database and static files from one environment to another (this command is also included in the pull request).  The usage of it would be:

```
$ fab sync_to:PROD --set FABENV=DEV
```

Fabric hosts can technically be a list of hosts, I don't think this will support that as is so if that should be supported it will take some more work.  If anyone has thoughts on the best way to do that let me know!

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -83,6 +83,7 @@ for more information on working with these:
 
 Environments
 ------------
+
 Multiple environments can be configured to work with a fabfile.  To do so, add
 additional dictionaries, like ``DEFAUL`` above to the ``FABRIC`` dictionary.
 For example::
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -58,16 +58,18 @@ module. Here's an example, that leverages some existing setting names::
   ALLOWED_HOSTS = ["example.com"]
 
   FABRIC = {
-      "DEPLOY_TOOL": "rsync",  # Deploy with "git", "hg", or "rsync"
-      "SSH_USER": "server_user",  # VPS SSH username
-      "HOSTS": ["123.123.123.123"],  # The IP address of your VPS
-      "DOMAINS": ALLOWED_HOSTS,  # Edit domains in ALLOWED_HOSTS
-      "REQUIREMENTS_PATH": "requirements.txt",  # Project's pip requirements
-      "LOCALE": "en_US.UTF-8",  # Should end with ".UTF-8"
-      "DB_PASS": "",  # Live database password
-      "ADMIN_PASS": "",  # Live admin user password
-      "SECRET_KEY": SECRET_KEY,
-      "NEVERCACHE_KEY": NEVERCACHE_KEY,
+      "DEFAULT": {
+          "DEPLOY_TOOL": "rsync",  # Deploy with "git", "hg", or "rsync"
+          "SSH_USER": "",  # VPS SSH username
+          "HOSTS": ["123.123.123.123"],  # The IP address of your VPS
+          "DOMAINS": ALLOWED_HOSTS,  # Edit domains in ALLOWED_HOSTS
+          "REQUIREMENTS_PATH": "requirements.txt",  # Project's pip requirements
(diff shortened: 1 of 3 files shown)


stephenmcd:
Only glanced at the code, but I like the idea.

joshcartme:
Cool, what do you think @jerivas, I think you had done the most recent work on the fabfile.

jerivas:
Nice idea! A few things:
- Are you going to update `local_settings.py.template` with the new format for the `FABRIC` dict?
- `GUNICORN_PORT` and `REPO_URL` were dropped, so no need to define them in the settings.
- I think you're missing a newline in line 88.

stephenmcd:
I think with the `sync_foo` commands, it'd be good to add a big red warning with a prompt to confirm, something like:

You're about to overwrite the database and uploaded media in FOO with database and uploaded media from BAR. Are you sure?

(where FOO and BAR are the env names in the `FABRIC` dict).

joshcartme:
Thanks for the feedback guys!

@jerivas I've updated the FABRIC dict in local_settings.py.template, removed GUNICORN_PORT from the example, and added the missing newline.

@stephenmcd I think that the confirm message was a really good idea and will save a lot of headaches for us and users in the future.  I've just added a confirm to the beginning of sync_to, the task that calls the other sync tasks.  Do you think there should be more than one confirm?  I also added a no_input argument to sync_to, I specifically tested `no_input != True` because fabric passes all arguments as strings.

I think it would be good to get some more people to test the sync_to command before we merge this.  It is working for me in a production environment, but that is using the old fabric file.

jerivas:
Oh, just remembered that you should update the docs too, both to 
document the new sync task and the new FABRIC syntax.

joshcartme:
Good call @jerivas, just updated the docs.  I'd appreciate it if a few people could give it a read through to make sure it makes sense.

Also, It would still be good for others to test the sync command.  Thanks!

jerivas:
Will try to give it a shot in the weekend.

PR #1343, opened 2015-07-03
Amend file location for local settings when advising on contribution.

sjdines:
The file `local_settings.py.template` does not exist in the directory `mezzanine/project_template/`. Change contribution advisory of the new location at `mezzanine/project_template/project_name`.


stephenmcd:
Thanks Stuart!

PR #1344, opened 2015-07-03
admin: use apps config verbose_name for display

gradel:
Works for apps that have no explicit app_config.
Will break with django < 1.7.


stephenmcd:
Nice, thanks!

PR #1349, opened 2015-07-08
Clean leading/trailing slashes from page admin slugs.

jxcl:
[`path_to_slug`](https://github.com/fusionbox/mezzanine/blob/7ca7a48ded1d831756995082f5803fce6211b2f1/mezzanine/utils/urls.py#L125-L137) strips leading and trailing slashes from a URL path. Allowing leading and trailing slashes in the `slug` field of a page will result in a page that will never be accessible. The leading and trailing slashes should be cleaned in the page form.

### DIFF
--- a/mezzanine/pages/admin.py
+++ b/mezzanine/pages/admin.py
@@ -11,7 +11,7 @@ from django.shortcuts import get_object_or_404
 from mezzanine.conf import settings
 from mezzanine.core.admin import DisplayableAdmin, DisplayableAdminForm
 from mezzanine.pages.models import Page, RichTextPage, Link
-from mezzanine.utils.urls import admin_url
+from mezzanine.utils.urls import admin_url, clean_slashes
 
 
 # Add extra fields for pages to the Displayable fields.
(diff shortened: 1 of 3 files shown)


Kniyl:
What happens to index pages in the page tree? They must have a slug of "/".

mezzanine/utils/urls.py
```diff
@@ -133,5 +133,13 @@ def path_to_slug(path):
     for prefix in (lang_code, settings.SITE_PREFIX, PAGES_SLUG):
         if prefix:
             path = path.replace(prefix, "", 1)
-    path = path.strip("/") if settings.APPEND_SLASH else path.lstrip("/")
+    path = clean_slashes(path)
     return path or "/"
```
stephenmcd: `return path or "/"` here should move to `clean_slashes` - that will allow for a homepage slug (eg "/") to be set from admin

mezzanine/pages/admin.py
```diff
@@ -27,10 +27,11 @@ class PageAdminForm(DisplayableAdminForm):
     def clean_slug(self):
         """
         Save the old slug to be used later in PageAdmin.save_model()
-        to make the slug change propagate down the page tree.
+        to make the slug change propagate down the page tree, and clean
+        leading and trailing slashes which are added on elsewhere.
         """
         self.instance._old_slug = self.instance.slug
-        return self.cleaned_data['slug']
+        return clean_slashes(self.cleaned_data['slug'])
```
stephenmcd: Should only run `clean_slashes` here if the model isn't the Link model, something like `if not isinstance(self.instance, Link)`

stephenmcd:
I left some comments on fixing this for homepages and Link pages, otherwise looks good.

Should also add some tests if possible.

jxcl:
@stephenmcd I have made the suggested changes.

stephenmcd:
Just tested this - it no longer auto-generates slugs from titles, which is a major feature.

Haven't dug into why, but this is broken ATM.

stephenmcd:
Actually I think it might be due to the feedback I gave, sorry :-) 

I'll dig into it.

stephenmcd:
Merged now, thanks guys!

Issue #1327, opened 2015-06-08
Admin's "View site" link doesn't work with multiple sites


simenheg:
In the admin interface, the "View Site" link isn't sensitive to changes to the current site.

For instance, imagine that two sites are set up under the same Mezzanine installation: `a.foo.org` and `b.foo.org`. Consider an admin logging in at `a.foo.org/admin`. After a while she decides to work on the sister site, `b.foo.org`, and uses the dropdown menu provided in the admin header to switch sites. Now the "View site" link is broken, since it still points to `a.foo.org`.

stephenmcd:
That's by design - the correct content is still loaded, and the correct site will still be identified (via admin user's session). It's just that the correct domain is not used, as per the admin site itself. This is to ensure all admin access is bound to the same domain as the admin site, which may be required for security.

simenheg:
Ah, that sounds right. Thanks for the quick answer!

There is still one thing that seems strange though, and I'm wondering whether that's by design as well, or if I'm doing something wrong. The loading of the `index.html` template still seems to be bound to the domain name. So in the example above, after switching from `a.foo.org` to `b.foo.org` in admin and clicking the "View site" button, you're taken to `a.foo.org` as expected. `b.foo.org` is set as the current site in the session, so the content for that site is loaded, but `index.html` is still fetched from `a/templates/index.html` instead of `b/templates/index.html`.

stephenmcd:
You're right, that's a bug in the host-based template loader. It looks at the domain and not the actual current site.

PR #1350, opened 2015-07-08
Make "View site" link to selected site.

JulianAndrews:
Currently if you select a different site from the sites dropdown and click "View site" it will take you to the site you're visiting the admin through rather than the selected site. This fix should make the link correctly bring you to the selected site instead.


stephenmcd:
Don't think this is required. You get the same domain you're already on, but content should be for the selected site. See #1327 for more info.

PR #1352, opened 2015-07-14
Fix #1351 - exception in displayable_links_js

### LINKED ISSUE #1351: Link dialog sometimes won't open in admin interface
In some situations, trying to open `/displayable_links.js` generates a 500 with the following message:

> TypeError at /displayable_links.js - unorderable types: dict() < dict()

https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/views.py#L197

Replacing this line with e.g. `return HttpResponse(dumps([link[1] for link in sorted(links, key=lambda link: link[1]['title'])]))` fixes the problem.

I'm using Python 3.4, but this line of code looks unintended and shouldn't work on Python 2 either :-)

AlexHill:
Links were being ordered by their dict representation, which I would yield unpredictable results in Python 2 and cause an exception in Python 3. Changed to order by the `'value'` key of the dict.

### DIFF
--- a/mezzanine/core/views.py
+++ b/mezzanine/core/views.py
@@ -194,7 +194,8 @@ def displayable_links_js(request):
             verbose_name = _("Page") if page else obj._meta.verbose_name
             title = "%s: %s" % (verbose_name, title)
         links.append((not page and real, {"title": str(title), "value": url}))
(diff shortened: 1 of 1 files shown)


stephenmcd:
Nice one Alex.

PR #1369, opened 2015-07-28
Proxy model doesn't show up in the "Add..." dropdown

acatton:
I have a proxy model with a different Admin, but I don't see it in the "Add..." dropdown.

This fixes the bug. (I added a style fix commit, if you don't like it, let me know, I can remove the second commit)


stephenmcd:
Thanks a lot Antoine.

PR #1372, opened 2015-08-03
Do not throw away next parameter on login

jxcl:
get_full_path() throws away the ``next`` URL parameter, breaking
vanilla Django redirection behavior. The login form should redirect
to the ``next`` parameter on a successful login.

I would be willing to go farther and say that when the ``next`` parameter is present then the Admin and Site buttons should not be present and the page should unconditionally redirect to the next destination.

### DIFF
--- a/mezzanine/core/middleware.py
+++ b/mezzanine/core/middleware.py
@@ -59,7 +59,7 @@ class AdminLoginInterfaceSelectorMiddleware(object):
             response = view_func(request, *view_args, **view_kwargs)
             if request.user.is_authenticated():
                 if login_type == "admin":
-                    next = request.get_full_path()
+                    next = next_url(request) or request.get_full_path()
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks Alexander.

BTW if you're curious, the code represents older versions of Django admin which didn't use a next param and would show the login page on any URL when unauthenticated.

jxcl:
@stephenmcd  Would you be amenable to a further cleanup of this function? It seems to me that because of this middleware, [the view is being called twice.](https://github.com/fusionbox/mezzanine/commit/bb08e42a740fe358d4718312e33847cafae0428b#diff-be6e2f8df43a1d10382305c000cedd75L59) It also seems to me that if no redirect is required, the original response should be returned, instead of creating a new but identical response.

stephenmcd:
Sure, go for it.

PR #1374, opened 2015-08-06
Allow all styles used by default tinymce toolbar.

ryneeverett:
- Add 'text-decoration' for Underline and Strikethrough.
- Add 'padding-left' for Indent.
- Add 'list-style-type' for non-Default Lists.
- Alphabetize for consistency with other settings.


stephenmcd:
Looks good, thanks Ryne.

PR #1379, opened 2015-08-14
Remove extra SitePermissionInline's. Fix #1366.

### LINKED ISSUE #1366: seeing duplicate inlines for site permissions using default mezzanine setup
1. open admin
2. create new user
3. scroll down to add site permissions and see the following screenshot
https://www.evernote.com/l/AfdtYsctJHVOSq5HO7xHIFEBrOcASiYHSqAB/image.png

here is my pip freeze:

````
Django==1.8.3
Fabric==1.8.3
Mezzanine==4.0.0
Pillow==2.9.0
Pygments==1.6
South==0.8.2
amqp==1.4.5
anyjson==0.3.3
argparse==1.2.1
astroid==1.1.1
beautifulsoup4==4.4.0
billiard==3.3.0.20
bleach==1.4.1
boto==2.38.0
celery==3.1.18
chardet==2.3.0
coverage==3.7.1
dj-database-url==0.3.0
django-contrib-comments==1.6.1
django-debug-toolbar==1.3.0
django-extensions==1.5.5
django-grappelli==2.6.1
django-require==1.0.7
django-storages==1.1.8
django-tastypie==0.10.0
django-wysiwyg-redactor==0.4.5.1
djangorestframework==3.1.3
ecdsa==0.11
fancycompleter==0.4
filebrowser-safe==0.4.0
future==0.14.3
grappelli-safe==0.4.0
gunicorn==19.3.0
html5lib==0.999999
kombu==3.0.26
logilab-common==0.61.0
mimeparse==0.1.3
oauthlib==1.0.0
paramiko==1.12.4
pdbpp==0.7
psycopg2==2.6
pycrypto==2.6.1
pylint==1.2.1
pyrepl==0.8.4
python-dateutil==2.2
pytz==2015.4
pyuploadcare==1.2.14
requests==2.7.0
requests-oauthlib==0.5.0
six==1.9.0
sqlparse==0.1.11
tzlocal==1.2
virtualenv==1.11.6
wmctrl==0.1
wsgiref==0.1.2
```

I also went into mezzanine source to check admin settings for inline model and found they are configured correctly at first glance:
https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/admin.py#L295-L298

ryneeverett:
Extra defaults to max_num (or 3, whichever is lesser) if extra is not set. In this case max_num was 1 so we were getting 1 extra inline.

### DIFF
--- a/mezzanine/core/admin.py
+++ b/mezzanine/core/admin.py
@@ -297,6 +297,7 @@ class SingletonAdmin(admin.ModelAdmin):
 
 class SitePermissionInline(admin.TabularInline):
     model = SitePermission
+    extra = 0
     max_num = 1
     can_delete = False
 
--- a/mezzanine/core/templates/admin/base_site.html
+++ b/mezzanine/core/templates/admin/base_site.html
@@ -36,7 +36,6 @@
 <script>
 jQuery(function($) {
     $('.admin-title').click(function() {location = window.__admin_url;});
-    $("#id_sitepermissions-__prefix__-sites").parent().parent().parent().remove();
 });
 </script>
 


stephenmcd:
Hey Ryne, thanks for this - I'm keen to properly fix it without the JS hack, but I tried your patch and it doesn't work quite right. 

With the patch and an existing user, I still see two inlines. With a new user, I get opne inline, but the changes don't persist - it's actually the invalid one (for lack of a better term) that gets removed by the JS hack. 

I guess it's related to some recent changes in Django admin - if you look at the form field that the existing JS hack removes, and that remains with your patch, its actual field name contains the string `__prefix__` within it - I think Django runs some JS which modifies that, and that's not running at the moment, but that's all a guess at this stage.

I'll leave this open for now since the code changed in it is probably partially correct, but it's incomplete.

ryneeverett:
> I think Django runs some JS which modifies that, and that's not running at the moment, but that's all a guess at this stage.

Good guess. It's not running because grappelli-safe overrides the `tabular.html` template which would otherwise call the relevant JS.

ryneeverett:
Wrong. See [grappelli-safe 56](https://github.com/stephenmcd/grappelli-safe/pull/56).

ryneeverett:
Of course, the js hack still ought to be removed.

PR #1382, opened 2015-08-19
Fixed smashed editable textarea.

ryneeverett:
Bootstrap sets `textarea {line-height: inherit;}` which smashes the
textarea.

I checked a few different form widgets and as far as I can tell setting
the line-height to zero only affects textarea's.

If it turns out what I removed is necessary, we could alternatively add
`.editable-form p > textarea {line-height: normal;}`.

smashed-textarea

### DIFF
--- a/mezzanine/core/static/mezzanine/css/editable.css
+++ b/mezzanine/core/static/mezzanine/css/editable.css
@@ -43,7 +43,7 @@ a#editable-toolbar-toggle {border-left:0px !important; text-decoration:none;
 .editable-form select {width:auto;}
 .editable-form input[type="file"] {width: 300px;}
 .editable-form p {margin: .5em 0 !important; padding: 0 !important;
-    line-height: 0 !important; border: 0 !important;}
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks a lot for finding this and fixing it.

A lot of the styles in here don't look like they change anything, but are there to prevent the underlying site's styles from bleeding into the modal - I can't recall for certain, but this is probably one of them, especially given the `!important` declaration.

Given that, we should fix this explicitly by targeting textareas in the modal, which is exactly what your other suggestion would do - do you want to set up a new PR? I can just make the change myself if you like, up to you - let me know.

ryneeverett:
@stephenmcd Gotcha, go ahead and make the change, thanks!

stephenmcd:
Will do, thanks again.

ryneeverett:
BTW, pretty sure my alternate suggestion will also require an `!important`.

stephenmcd:
Got it working.

PR #1383, opened 2015-08-20
jQuery 1.8.3 and jQuery UI 1.8.24

nikolas:
### jQuery upgrade
This upgrades jQuery from 1.7.2 to 1.8.3, which is the latest in the 1.8 series.

### jQuery UI upgrade
This upgrades jQuery UI from 1.8.2 to 1.8.24, fixing
a bunch of bugs.

It also changes the CSS file used for jQuery UI - the v1.9
CSS was being used, so I've replaced it with the appropriate
CSS for version 1.8.24.


jerivas:
Hi @nikolas. Just to make sure everything is working as expected, have you checked the following...?

- Admin: site picker, language picker, admin actions dropdown
- Admin: collapsible admin panel
- Admin: sortable inlines
- Admin: modeltranslation language switcher (on individual model fields)
- Admin: collapsible fieldsets
- Frontend: all Bootstrap plugins (I suspect this works fine)
- Frontend: inline editor

Upgrading jQuery is tricky because the admin uses a lot of plugins. Perhaps this is a good opportunity to upgrade those too, and pave the way for the upcoming release of Bootstrap 4.

nikolas:
@jerivas I've added some commits, in one of them I've updated the jquery form plugin, because I found a new version for that. I also found a bug with the edit pop-up overlay being positioning at the bottom of the screen, but I've fixed that and it's now behaving just like it was previously.

Here's a list of the things I've confirmed are working. The stuff that's not tested I'm
still figuring out how to add to my site so I can test it (e.g. I haven't found the multi-language
features yet).

* Admin: site picker **works**, language picker, admin actions dropdown
* Admin: collapsible admin panel **works**
* Admin: sortable inlines **works**
* Admin: modeltranslation language switcher (on individual model fields)
* Admin: collapsible fieldsets **works**
* Frontend: all Bootstrap plugins (I suspect this works fine)
* Frontend: inline editor **works**
* Both: TinyMCE editor **works**

stephenmcd:
Amazingly this all works - the positioning of the front-end editing modals is all over the place though, so that'll need looking at before this can be merged.

stephenmcd:
Here are some screenshots, compare these to the current version:

screen shot 2015-08-21 at 10 23 16 am
screen shot 2015-08-21 at 10 23 24 am

Note that these include the bootstrap upgrade also, which on its own works fine (doesn't produce these layout issues)

jerivas:
Steve, do you have anything else in mind that might need to be checked? The sortable page tree comes to mind. I think we should come up with an exhaustive list of all the things that jQuery can break, as this will be a recurring task for the foreseeable future.

I've encountered the inline editor issue before, and @joshcartme showed me a solution for that. Simply update `static/mezzanine/js/jquery.tools.overlay.js` to [this](https://gist.github.com/jerivas/89dc9314aa651168999c). The actual issue appears to stem from [here](https://github.com/jquerytools/jquerytools/issues/1007).

IMO we should go even further and try to upgrade to the latest version in the 1.x branch, or even 2.x, which AFAIK is the same, it simply drops support for IE8. Bootstrap 4 is already in alpha state and they have dropped IE8 too.

stephenmcd:
> do you have anything else in mind that might need to be checked?

Nope that original list looks good

joshcartme:
@stephenmcd try updating jquery.tools.overlay.js to this, https://gist.github.com/anonymous/c6ce5a4887b41f58e32e

I don't remember exactly what I changed in that (something minor on like two lines) but something in jQuery around 1.8 changed that broke the bundled version.  I've used that quite a few times to fix the overlay position in projects that needed > the bundled jQuery.

Alternatively we may just be able to upgrade the version of jquery.tools.overlay

I'm heading out on vacation tomorrow so I won't be able to be responsive in this thread.

nikolas:
I've rebased this branch on the current master and fixed the overlay centering problems.

stephenmcd:
Thanks @joshcartme and @nikolas - I'll test the PR again.

stephenmcd:
Latest change fixes the positioning, but reduces width when TinyMCE is included, and now the toolbar wraps

stephenmcd:
I've got a fix for the toolbar wrapping so leave that with me.

Thanks a lot for this, it's a huge effort!

PR #1387, opened 2015-08-20
Test on Python 2.7, Django 1.8

nikolas:
These are the versions I use by default, so we shouldn't need to exclude this combination.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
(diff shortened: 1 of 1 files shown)


stephenmcd:
See 27b3d5d373ed3685b9c1db5640e36820625f70ad

stephenmcd:
Actually I rethought this and I'll merge it - it's the lesser of two evils really, better to have to manually update to point releases thank skipping a whole version because of bad packaging.

PR #1388, opened 2015-08-21
Bootstrap 3.2.0

nikolas:
Continuing the effort of gradually updating bootstrap,
this upgrades from 3.1.1 to 3.2.0.

There's not many changes here and everything's looking okay
to me.


stephenmcd:
Thanks! 

I guess this makes all the work and testing of the upgrade yesterday redundant - let's try and avoid that :-)

nikolas:
I wouldn't say it's redundant.. I just thought it was safer to upgrade slowly than all at once. If you'd like to avoid that though, when I upgrade jQuery again I can go straight to version 1.11.x instead of 1.9.x.

stephenmcd:
I'd say move straight to the latest version of something that doesn't break anything :-)

nikolas:
Alright, makes sense!

PR #1392, opened 2015-08-26
Update tinymce_setup.js: Django language codes

fygul:
https://docs.djangoproject.com/en/1.8/topics/i18n/
Language codes are generally represented in lower-case, but the HTTP Accept-Language header is case-insensitive. The separator is a dash.

I also noticed a warning message showing when I run 'python manage.py runserver'. it's about 'zh-tw' will be deprecated in Django 1.9, use 'zh-hant' instead. So I also add 'zh-hant'. I guess 'zh-cn' may be deprecated too.

I tested 'zh-hant' and 'zh-tw' well, my tinymce editor shows hints in Traditional Chinese now.


stephenmcd:
Awesome, thanks a lot!

stephenmcd:
What's your full name? I'll add it to the AUTHORS files.

Thanks again.

fygul:
Fygul Hether
Thanks.
You also remind me to update my profile of github account. ;-)

stephenmcd:
Great, thanks Fygul.

Issue #1395, opened 2015-08-29
EXTRA_MODEL_FIELDS cant refer to self in ForeignKey


ZeroXn:
i was trying to add a parent category reference to blogcategory, but each time got and error such as
    `blog.blogcategory: 'parent_category' has a relation with model self, which has either not been installed or is abstract.`
i had to modify boot/\_\_init\_\_.py as mention at the link below, to get it to work
https://groups.google.com/forum/#!topic/mezzanine-users/CBqb6srpmqU

Is this a bug? or is there some other way to implement without modifying 'boot/\_\_init\_\_.py' directly?

stephenmcd:
Sounds like you have some working code, please create a PR.

ZeroXn:
ok np, i submtting it as pr #1396

PR #1396, opened 2015-08-30
re-send class_prepared signal for self-referenced ForeignKeys

ZeroXn:
pr for issue #1395

### DIFF
--- a/mezzanine/boot/__init__.py
+++ b/mezzanine/boot/__init__.py
@@ -52,8 +52,13 @@ def add_extra_model_fields(sender, **kwargs):
     by the ``EXTRA_MODEL_FIELDS`` setting.
     """
     model_path = "%s.%s" % (sender.__module__, sender.__name__)
-    for field_name, field in fields.get(model_path, {}).items():
-        field.contribute_to_class(sender, field_name)
+    extra_fields = fields.get(model_path, {})
+    if (extra_fields):
+        for field_name, field in extra_fields.items():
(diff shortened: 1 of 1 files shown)


stephenmcd:
Can we add some comments explaining what the new lines do?

ZeroXn:
in a new pr? or just add them here?
but as i mentioned in my initial issue, it not my code, and i linked to the source.
The way i understand how it works is like this
    if (extra_fields):
prevents the class_prepared being fired unnecessarily

    del fields[model_path]
removes an already processed field

    class_prepared.send(sender=sender)
process any remaining fields

stephenmcd:
Re-sending `class_prepared` sounds wrong, wouldn't it potentially break other people's code if they relied on it only being sent once? Any thoughts @Kniyl ?

AlexHill:
I refactored a bunch of code in this area for Django 1.9, and can confirm that referring to `"self"` works fine in 1.9 under Mezzanine's `1.9_compat` branch (see #1481).

I don't mind taking a look at this after that's merged if somebody needs it for 1.8, but probably not before then.

stephenmcd:
I think having it fixed for 1.9 is good enough.

jeffrand:
This feature would be extremely useful for me, and I'm not able to upgrade to 1.9 in the near term. @stephenmcd, would you reconsider merging this?

stephenmcd:
@jeffrand your best bet would be to apply the change  to a fork of your own, and deploy against that.

PR #1398, opened 2015-09-01
Thumbnails generation with S3BotoStorage

dablak:
I'm using S3BotoStorage as the default file storage for media files.

When uploading a new image from the admin panel, this is stored in S3. Then a thumbnail is generated and stored locally and in S3, so we have 2 copies of the same thumbnail.

The problem happens when deleting the image and uploading it again. When deleting the image, the image and the thumbnail are deleted from S3 but not the local thumbnail. Then, when uploading the same image again, a new thumbnail is not being generated or uploaded to S3 because there is already a local thumbnail generated.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -319,21 +319,18 @@ def thumbnail(image_url, width, height, upscale=True, quality=95, left=.5,
     if image_url_path:
         thumb_url = "%s/%s" % (image_url_path, thumb_url)
 
-    try:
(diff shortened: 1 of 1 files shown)


stephenmcd:
The thumbnails are intentionally always local - making a ton of requests to S3 to check for thumbnails during one request to the site isn't appropriate.

Correct fix would be to ensure thumbnails correctly get removed locally.

dablak:
What if you have more than one instance of mezzanine running? In that case
you can't just remove locally the thumbnails in all of them.

IMO the right solution is to store in the database a thumbnail Imagefield.
So you don't need to hit S3 each time to know if it has been created.

PR #1401, opened 2015-09-06
Add support for blogs with > 500 posts.

annabunches:
This just puts the existing code in a loop and made multiple max-sized calls to the gdata API, with some tracking state to make sure all the posts get pulled.


stephenmcd:
Awesome, thanks a lot Anna.

PR #1404, opened 2015-09-09
Update tinymce_setup.js

fygul:
Add two often used buttons: removeformat (Clear formatting) and blockquote.


stephenmcd:
Gonna try and keep this minimal - you can always override it.

stephenmcd:
(thanks anyway!)

PR #1406, opened 2015-09-13
Add several new overridable blocks to the base template.

annabunches:
New blocks:

* a navbar_title block, to make it easy to replace the site title with an image logo, for instance.
* a navbar_search_form block. I found that I didn't want the "all" dropdown choices, and this makes it easy to invoke your own search_form tag.
* and a navbar_dropdown_menu block, in case you need to change that behavior (or just remove it) as well.


stephenmcd:
Thanks Anna!

PR #1409, opened 2015-09-14
Fix and test page processor's exact_page argument

dsanders11:
It appears that the `exact_page` argument for page processors broke somewhere around commit a28cdf4 with the removal of the `page.set_helpers` call. This leads to any usage of the `exact_page` argument causing an error in `PageMiddleware` (see build failure for stephenmcd/cartridge#265).

Added a test which confirms it was broken, and fixed by using the page context processor to ensure `is_current` is set correctly during `PageMiddleware`.


stephenmcd:
Nice one, thanks.

PR #1410, opened 2015-09-15
Disqus sso encode fix

pirave:
SSO did not work with the current encoding for the payload. Using str.encode('utf8') seemed to fix the issue.


pirave:
The first two commits are for a different issue.

stephenmcd:
Thanks

PR #1411, opened 2015-09-16
Resolve css/js Media paths using static templatetag

dsanders11:
When running Mezzanine with a CDN for CSS/JavaScript (or even just browser side caching) it's important to use some form of cache-busting names for static files like those generated by CachedStaticFileStorage.

Static files listed under `class Media:` do not automatically get these benefits, so it is required to manually convert them to their static file names. Django's admin does this in several places.

Otherwise caching (via CDN or browser) can wreak some havoc when these files change and old versions are used.


stephenmcd:
Makes sense, thanks.

PR #1412, opened 2015-09-18
Implemented InMemoryUploadedFile method for thumbnail generation. 

renyi:
Allows for pure remote thumbnail hosting. I only tested this on my limited scenarios (S3 boto). It works for Media Library (slow... list_dir() from filebrowser needs to be reimplemented) and templates.

However, it's only limited to MEDIA_URL. If you have any {{ STATIC_URL }}{% thumbnail "image.jpg" %}, this will fail.

Feedbacks to improve this are welcomed. Thanks, :D


stephenmcd:
Nice idea, don't think we'll introduce it though - perhaps better off as a separate app containing the alternate template tag.

Issue #1413, opened 2015-09-24
HTMLParseError disappeared in Python 3.5


bors-ltd:
Hi,

I was surprised by a Python 2 import error on Mezzanine start. It happened it was an ImportError when trying to import HTMLParseError that triggered the Python 2 code path.

HTMLParseError just disappeared in Python 3.5, the parser does not fail on bad HTML any more: https://docs.python.org/3.5/whatsnew/3.5.html#changes-in-the-python-api

(The release notes mention HTMLParserError but that's a typo.)

I came out with the following patch:

```diff
diff --git a/mezzanine/utils/html.py b/mezzanine/utils/html.py
index fd5bd10..4d3c765 100644
--- a/mezzanine/utils/html.py
+++ b/mezzanine/utils/html.py
@@ -2,8 +2,13 @@ from __future__ import absolute_import, unicode_literals
 from future.builtins import chr, int, str

 try:
-    from html.parser import HTMLParser, HTMLParseError
+    from html.parser import HTMLParser
     from html.entities import name2codepoint
+    try:
+        from html.parser import HTMLParseError
+    except ImportError:  # Python 3.5+
+        class HTMLParseError(Exception):
+            pass
 except ImportError:  # Python 2
     from HTMLParser import HTMLParser, HTMLParseError
     from htmlentitydefs import name2codepoint
```
My Mezzanine instance was starting again, but it's not tested thoroughly.

stephenmcd:
Fix looks fine - could you create a pull request with it? Thanks a lot.

bors-ltd:
Against a specific branch, or master?

I was on the 4.0.1 branch for this (still working on my swappable BlogPost).

stephenmcd:
280642c8d0e7e41b90398c974a15533f6df20049

PR #1414, opened 2015-09-24
utils.html: HTMLParseError disappeared in Python 3.5

bors-ltd:
The HTMLParser is guaranteed not to choke on HTML soup.

Following up issue #1413


stephenmcd:
Thanks a lot!

PR #1416, opened 2015-09-29
Adding Google recaptcha option to forms

eino-makitalo:
I added google recaptcha as an option to forms.  It's not used by default, so it should not change user experience.  Was not able to test everything my  Windows environment, but tried to ensure python 2/3.


stephenmcd:
Thanks a lot for this! 

I think it would be better implemented as a plugin. Have a look at this `FORMS_EXTRA_FIELDS` setting, that should allow you to include a custom field type that embeds the captcha and performs validation.

eino-makitalo:
Thx.
I'll check this FORMS_EXTRA_FIELDS.  I feel that captcha is not field in that sense that we do not want to perform validation in that sense that we will give error message to robotspammer. It's also field added only once into form.  It's not like email or phonenumber field.  But adding it like extra field type you can choose position of that field.  In any case it's good to be near submit button. 

Plugin is OK I think but without captcha processing forms are useless unless authenticated and that was my motivation to add functionality to existing module directly.

PR #1421, opened 2015-10-04
Add kwarg for form used in login and password reset views

dsanders11:
Same change was made in Cartridge a good while back, makes it easier to change the form to a custom one.


stephenmcd:
Thanks David

PR #1422, opened 2015-10-04
Don't titlecase group name for ADMIN_MENU_ORDER

dsanders11:
Since the group title is user-specified it is counterintuitive to apply titlecase to it since it will display differently than what was specified in settings.


stephenmcd:
Thanks!

PR #1423, opened 2015-10-06
Only use old_account_redirect if ACCOUNT_URL changed

dsanders11:
Right now `old_account_redirect` will cause an infinite redirect loop if `ACCOUNT_URL` is set to `OLD_ACCOUNT_URL` and a Page under `ACCOUNT_URL` is accessed.

Patch breaks PEP8 for line length but couldn't see a clean way to break it so deferring to @stephenmcd's preference.


stephenmcd:
I think we can just remove `OLD_ACCOUNT_URL` handling altogether - it's only there for backward compat but it's been there for a long time now. I'll push it shortly.

Thanks

Issue #1154, opened 2014-11-13
page_menu templatetag optimization


duduklein:
In the page_menu template tag, we check if the "context_menu" is already in the context, so we don't make another call to the db (this is an expensive call with many "left outer joins").

Suppose we have a base template with 2 blocks (header and footer) and another one that extends the base like this:

```
{% extends "base.html" %}
{% load page_tags %}

{% block header %} {% page_menu template_name1 %} {% endblock %}
{% block footer %} {% page_menu template_name2 %} {% endblock %}
```

In this case, we'll do 2 different queries to the db, because when we add an entry to the context, this entry is only available in the block, as stated [here](https://docs.djangoproject.com/en/dev/howto/custom-template-tags/#setting-a-variable-in-the-context).

One possible solution to this issue is doing `context.dicts[0]["menu_pages"] = pages` instead of `context["menu_pages"] = pages`. I'm not sure this is the best solution, since I'm not sure we are supposed to access all the "context" dictionaries directly.

One other solution is to create a templatetag that does not render any html, but only adds this variable to the context. The developer that needs it would call it in the base template (main block), so this variable would be available to all other blocks.

What do you think?

PR #1424, opened 2015-10-06
Add _parent_page_ids to global context scope

julianandrews:
Currently if the first menu loaded is in a nested context
` parent_pages_ids` can drop out of scope, and then never get reset since
`menu_pages` is set in `context.dicts[0]`.

See https://github.com/stephenmcd/mezzanine/issues/1154.


stephenmcd:
Makes sense - I guess the same logic should apply to all the context variables in this section, but I'll fix that up.

Thanks

PR #1427, opened 2015-10-07
Upgrade jQuery, UI, and jQuery mobile

nikolas:
jQuery version 1.9.1
jQuery UI 1.10.4
jQuery mobile 1.3.2


stephenmcd:
Breaks page re-ordering on chrome and safari at least.

As discussed previously, these front-end upgrades are definitely appreciated, but need significant testing - please don't submit them without it.

nikolas:
Sorry.. my testing was incomplete. I'll re-open when that problem is addressed.

PR #1428, opened 2015-10-07
tinymce: add simple email plugin

gradel:
copied from https://gist.github.com/Melros/9119246


gradel:
Looks like tinymce does insert the mailto link already.

PR #1430, opened 2015-10-08
Don't overwrite files options in mezzanine-project command

nikolas:
Updating the `files` option here disables the `--name` option, which I need
to render custom template files (i.e. a Makefile) when making a custom
mezzanine template for use with `mezzanine-project`.

This change adds `local_settings.py.template` to the list of files to render instead
of overwriting the list completely. This allows users to specify their own template
files if necessary.

Relevant django code is here: https://github.com/django/django/blob/master/django/core/management/templates.py#L56

And my mezzanine template I'm getting this to work with is here: https://github.com/nikolas/ctlmezzanine


stephenmcd:
Makes sense - thanks!

PR #1431, opened 2015-10-09
Crop parameter added for the thumbnail template tag.

winnukem:
Sometimes there is a need to resize an image to maximum width and height while keeping its ratio. I think it's a default behaviour in sorl-thumbnail, and I haven't found a way to do it in Mezzanine without padding, so I added this parameter.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -264,14 +264,17 @@ def search_form(context, search_model_names=None):
 
 @register.simple_tag
 def thumbnail(image_url, width, height, upscale=True, quality=95, left=.5,
-              top=.5, padding=False, padding_color="#fff"):
+              top=.5, padding=False, padding_color="#fff", crop=True):
     """
     Given the URL to an image, resizes the image using the given width
     and height on the first time it is requested, and returns the URL
-    to the new resized image. If width or height are zero then original
-    ratio is maintained. When ``upscale`` is False, images smaller than
-    the given size will not be grown to fill that size. The given width
-    and height thus act as maximum dimensions.
+    to the new resized image. If width or height are zero, the original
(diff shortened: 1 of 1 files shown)


joshcartme:
Hi @winnukem, you can already accomplish the same thing by passing zero as the width or height of the thumbnail.  For example:

    {% thumbnail image.file 500 0 %}

Would create a thumbnail with a width of 500 and height that maintained the aspect ratio.

winnukem:
Hi @joshcartme. How do I create a thumbnail with maximum width 500 and maximum height 500 while keeping its aspect ratio?

joshcartme:
Sorry about that @winnukem, I misunderstood the intention of the PR, I'll reopen and let other people weigh in.

stephenmcd:
If you specify a width and height cropping should already occur.

winnukem:
@stephenmcd yes, that is the only behaviour now, and this option lets you change it so that the image is NOT cropped

stephenmcd:
Sorry if I'm missing something, but I don't understand what you mean by "resize an image to maximum width and height while keeping its ratio" without cropping or padding.

winnukem:
For example, with
`{% thumbnail image.file 200 100 crop=False %}`
I want the following thumbnails to be generated:
* for 400x400 image - 100x100 thumbnail
* for 200x400 image - 50x100 thumbnail
* for 400x200 image - 200x100 thumbnail

That is, thumbnail size is limited by given bounding box, but one of its dimensions may be smaller so that the original image's aspect ratio is preserved.

stephenmcd:
That sort of makes sense, I think it's very confusing though.

winnukem:
Well, I just didn't expect thumbnail tag to crop my pictures by default without having a way to disable it (my pull request only gives you this ability and doesn't break any of the code that relies on current default behaviour). Also, as I have already mentioned, this is the default behaviour in sorl-thumbnail, which seems to be the most popular Django thumbnail app.

PR #1432, opened 2015-10-10
Switch to a lazy static for Media statics

dsanders11:
This is a mulligan on #1411. After more use, it seems there's a catch-22 situation with the current version. If `DEBUG` is set to `True` then the staticfiles storage just returns the same value it got, but if `DEBUG` is set to `False`, it tries to find the file in `STATIC_ROOT` (at least for the ManifestStaticFilesStorage storage). Since the files will only be in `STATIC_ROOT` if you ran `collectstatics`, you end up not being able to run `collectstatics` with `DEBUG` disabled because Django trips over the `static(...)` calls when loading the models before running the management command.

Long way of saying, lazy version solves this problem and may even be a bit better performing.

### DIFF
--- a/mezzanine/conf/admin.py
+++ b/mezzanine/conf/admin.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 from copy import copy
 
 from django.contrib import admin
-from django.contrib.admin.templatetags.admin_static import static
 from django.contrib.messages import info
 from django.http import HttpResponseRedirect
 from django.utils.translation import ugettext_lazy as _
@@ -13,6 +12,7 @@ from mezzanine.core.admin import BaseTranslationModelAdmin
 from mezzanine.conf import settings
 from mezzanine.conf.models import Setting
 from mezzanine.conf.forms import SettingsForm
+from mezzanine.utils.static import static_lazy as static
 from mezzanine.utils.urls import admin_url
 
 
--- a/mezzanine/core/admin.py
+++ b/mezzanine/core/admin.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 from django.contrib import admin
-from django.contrib.admin.templatetags.admin_static import static
 from django.contrib.auth import get_user_model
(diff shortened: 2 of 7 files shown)


stephenmcd:
Makes sense, thanks a lot David.

stephenmcd:
Hey David, this actually broke when using modeltranslation

I've made the following patch but you should check it to ensure it's working as expected with your changes:

a94e1948c6b0e98ddb1f6a813c35ff9fbdeb2b06

You provide a lot of contributions to the project and that's great - but I would recommend for developing, I'd work with all of the major features turned on in local_settings.py - modeltranslation and the accounts app are the main ones that come to mind. 

Thanks again!

dsanders11:
Sorry about that, didn't trigger a build failure so I figured it was alright. Tested a94e194 and it looks good to me on my end.

I'll try to be tighter with testing contributions in the future. Many of them are quick fixes during my own development and I don't have time to do extensive testing on Mezzanine vanilla where tests don't catch issues.

Would it be possible to get a development project similar to `project_template` that has those features turned on (and a `requirements.txt` which has the needed dependencies) to ease the process?

stephenmcd:
No sweat and thanks for testing my change.

Honestly I think adding modeltranslation and mezzanine.accounts are the only things needed right now that aren't tested automatically.

dsanders11:
Alright, I'll keep those two in mind in the future then.

stephenmcd:
FWIW here's what I have in local_settings to enable these:

LANGUAGE_CODE = "en"

from django.utils.translation import ugettext_lazy as _

```
USE_I18N = USE_L10N = USE_MODELTRANSLATION = True

# Supported languages
LANGUAGES = (
    ('en', _('English')),
    ('es', _('Spanish')),
)

INSTALLED_APPS += ("mezzanine.accounts",)
```

Actually I think accounts gets added during official tests, it's just modeltranslation that does. All the same, it's good to manually test with these things enabled even if the build passes - it definitely won't always catch everything.

PR #1433, opened 2015-10-12
In the admin dropdown menu, show add links to users without change permissions

Pieter1024:
This is a patch to fix the following problem:

1. Create a user with add permissions but not change permissions on a model
   that is managed through the admin (e.g. blog posts).
2. Log in as that user in the admin site.
3. Click on the name of the model in the dropdown menu on the left.
4. Error.

The problem arises because the template uses the admin_url instead of the
add_url for users without appropriate permissions to change instances of the
model. It then tries to remedy the situation by manually appending 'add/' to
the url. However, the url it receives from the template tag admin_url is set
to 'None'. The result is that the link points to '/admin/Noneadd'.


stephenmcd:
Thanks! Can you let me know your full name so I can add it to the contributors list?

PR #1436, opened 2015-10-16
Fix for is_spam_akismet to handle Python 3's bytestring response properly

xnx:
I found these changes make spam-checking with Akismet work properly on Python 3. The problem is to do with the different way that Python 3's urllib and Python 2's urllib2 deal with strings.


stephenmcd:
Awesome, thank you!

PR #1437, opened 2015-10-18
Replacement CommonMiddleware to handle Page URLs

dsanders11:
NOTE: This is not a ready to merge pull request, it's mainly to start a conversation.

Ran into this issue a few times, it's a rather large PITA. Basically (if `APPEND_SLASH=True`), since the `mezzanine.pages` urlconf has a pattern that matches all possible URLs, the Django `CommonMiddleware` will consider any invalid URL path missing a final slash (so any non-existent filenames, for example) to need a redirect to the URL with a final slash since it matches a URL pattern.

Unfortunately `CommonMiddleware` is overzealous and issues permanent redirects by default. So, for cases where an invalid filename is used, `CommonMiddleware` will issue a permanent redirect to the filename with a final slash appended. With CDNs or just browser caching, this is quite problematic, as those who see and cache the permanent redirect won't see that filename if it ever becomes valid. There's a variety of cases where that's problematic: 1) filenames following a pattern (date in name), someone jumps the gun and tries to see the next one in the sequence that isn't there yet, won't be able to see it when it is. 2) upload a file with a typo in the name, attempt to go to the correct filename, get a permanent redirect, fix filename typo, can't see file.

The best solution I could come up with involves a drop-in replacement for the `CommonMiddleware`. Unfortunately there's no easy hook for that particular bit of code so it leads to a copy-and-paste of most of the middleware. This pull request has the new logic in `is_valid_path` and hooks in at L357.

### DIFF
--- a/mezzanine/core/middleware.py
+++ b/mezzanine/core/middleware.py
@@ -1,17 +1,24 @@
 from __future__ import unicode_literals
 from future.utils import native_str
 
+import logging
+
+from django import http
 from django.contrib import admin
 from django.contrib.auth import logout
 from django.contrib.messages import error
 from django.contrib.redirects.models import Redirect
 from django.core.exceptions import MiddlewareNotUsed
-from django.core.urlresolvers import reverse, resolve
+from django.core import urlresolvers
+from django.core.urlresolvers import reverse, resolve, Resolver404
 from django.http import (HttpResponse, HttpResponseRedirect,
                          HttpResponsePermanentRedirect, HttpResponseGone)
+from django.middleware.common import CommonMiddleware as DjangoCommonMiddleware
 from django.middleware.csrf import CsrfViewMiddleware, get_token
 from django.template import Template, RequestContext
+from django.utils import six
 from django.utils.cache import get_max_age
+from django.utils.http import urlquote
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
 
@@ -19,12 +26,12 @@ from mezzanine.conf import settings
 from mezzanine.core.models import SitePermission
 from mezzanine.core.management.commands.createdb import (DEFAULT_USERNAME,
                                                          DEFAULT_PASSWORD)
+from mezzanine.pages.models import Page
 from mezzanine.utils.cache import (cache_key_prefix, nevercache_token,
                                    cache_get, cache_set, cache_installed)
 from mezzanine.utils.device import templates_for_device
 from mezzanine.utils.sites import current_site_id, templates_for_host
 from mezzanine.utils.urls import next_url
-
 
 _deprecated = {
     "AdminLoginInterfaceSelector": "AdminLoginInterfaceSelectorMiddleware",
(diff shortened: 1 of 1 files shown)


stephenmcd:
Not keen on this at all. That's a bunch of code we now have to maintain in order to handle arguably a bunch of edge cases.

Why not file this as a bug with Django?

dsanders11:
It's a lot easier to override the behavior in [Django 1.9](https://github.com/django/django/blob/stable/1.9.x/django/middleware/common.py#L68) but that's still months away from being released, and I assume Mezzanine will continue to support 1.7 and 1.8 for a while. Doing a quick read over of the Django backport policy it's highly unlikely they would backport that change to 1.8.

So, with the 1.9 version there would only need to be around 25 lines of code in this PR compared to the 100 currently. Would that be an acceptable amount of code for this drop-in middleware? Or would even that be too much?

My main concern is that the current behavior is unintuitive with the redirects to non-existent filenames with an appended slash, adds an unnecessary redirect on 404s, and can in some edge cases be dangerous with the permanent redirects (in ways that less technical users are likely to be hard pressed to figure out).

The proposed change brings Mezzanine into line with how most major websites behave with regards to this: true 404s don't append a slash, existing pages do append a slash.

AlexHill:
Hey David, thanks for submitting this.

> My main concern is that the current behavior is unintuitive with the redirects to non-existent filenames with an appended slash, adds an unnecessary redirect on 404s, and can in some edge cases be dangerous with the permanent redirects (in ways that less technical users are likely to be hard pressed to figure out).

Forgive me if I'm missing something, but doesn't your PR only address this for `Page`s, where this isn't really a problem? On the other hand, the issue with uploaded filenames is annoying, but I'm not getting how this addresses it. Can you walk me through it?

Overall I agree with Steve here, this is a general Django gripe so ideally any energy spent trying to solve it would be best spent trying to solve it for everyone. A simple improvement might be to not append slashes to URLs ending in known filetypes (there's a list in the standard library's `mimetypes` module).

dsanders11:
@AlexHill 

> Forgive me if I'm missing something, but doesn't your PR only address this for Pages, where this isn't really a problem? On the other hand, the issue with uploaded filenames is annoying, but I don't get how this code addresses it. Can you walk me through it?

The main issue is the catch-all nature of the pages URL pattern in `mezzanine.pages`. Since that pattern will catch every possible URL (it has to in order to perform the `Page` lookup), `CommonMiddleware` always thinks there is a valid URL to redirect to with a slash. So on a 404 for a URL without a trailing slash (filename) `CommonMiddleware` checks to see if the URL with a trailing slash is 'valid' (which will always be true due to that URL pattern for `mezzanine.pages`) and then issues a permanent redirect. In reality if the attempted URL was a filename, the URL with a trailing slash almost certainly doesn't exist (www.foo.com/bar.jpg -> 404 -> www.foo.com/bar.jpg/ -> 404). Since `CommonMiddleware` by default issues a permanent redirect, a browser (or CDN trying to find the JPG being accessed) can cache the redirect and never be able to see the intended filename, even if it is valid in the future.

As mentioned in the original comment, there's a myriad of situations this can be problematic. Uploading an image, sending out the link, realizing there's a problem with the image, deleting it until the issue is fixed (wrong price for an ecommerce ad for example, needs to be addressed immediately), and then uploading the new image with that same name. If you did an email blast and then took 5 minutes between deleting the image and replacing it with a new one, then anyone who tried to view that image in those 5 minutes will get the permanent redirect and may never be able to see the image.

Another example would be forgetting to run `collectstatic` on deployment to production (or having it delayed) which is where I've seen it be a real PITA. The generated HTML will have links to filenames that will 404 since the files weren't deployed, and then the browser will cache a permanent redirect for those filenames.

This PR fixes the issue by having the replacement middleware not assume any URL which matches a pattern is valid. It has knowledge of how `Page` works in Mezzanine and checks to see if the `Page` actually exists for that URL before issuing a redirect.

As mentioned in my second comment here, this is clearly a known stumbling block to the Django developers since they added a new method ([should_redirect_with_slash](https://github.com/django/django/blob/stable/1.9.x/django/middleware/common.py#L69)) in 1.9 to make it easier to override this behavior without the copy and paste this PR involves.

Even with the easier overriding in 1.9, there would still need to be a replacement middleware and a suggestion to users to use it, or some sort of dynamic injection into the `should_redirect_with_slash` method.

I've been running this middleware in production for a few months and I have *significantly* less headaches due to issues with bad URLs due to the default behavior.

AlexHill:
Hi David,

Thanks for going into detail. I'm confident that I understand what the code does.

I think the reason I didn't really get this is that in my sites, the page handler is never going to be handling static/media file URLs in production - those prefixes are entirely handled by nginx or similar.

In development they're served by the `staticfiles` app. I didn't realise that `staticfiles` doesn't use the url resolving machinery at all, with the result that static files URLs end up matching the pages URL pattern when they're checked in `CommonMiddleware`.

I'm convinced that this is worth addressing. Not a fan of the middleware approach pre-1.9 though.

Here's some relevant discussion on URL fallthrough in Django: https://groups.google.com/d/topic/django-developers/64I3Qy4OH-A/discussion

That leads to https://github.com/raiderrobert/django-multiurl, which we may be able to use or adapt to solve this problem without modifying middleware.

AlexHill:
> I didn't realise that `staticfiles` doesn't use the url resolving machinery at all, with the result that static files URLs end up matching the pages URL pattern when they're checked in `CommonMiddleware`.

Oops, this is wrong. `staticfiles` skips middleware entirely when it's serving files.

dsanders11:
> I think the reason I didn't really get this is that in my sites, the page handler is never going to be handling static/media file URLs in production - those prefixes are entirely handled by nginx or similar

You're right, a carefully configured web server can avoid this by only serving certain prefixes and not forwarding calls to Django. I have a setup where if the file exists it is served, otherwise it is sent to Django. This allows for situations where there's generated content mixed in under the static/media prefix, say if you wanted to move sitemap.xml to under the media prefix or have a view under there such as www.foo.com/media/event_calendar.txt.

If you're aware of the pitfalls (which are not abundantly clear when you first hit them) then you can avoid the issue at the small expense of total flexibility in how you lay out URLs.

> I'm convinced that this is worth addressing.

Great. In case there's any other question of if it is worth addressing after the point about web server set up avoiding those paths, I'll reiterate my earlier points that best practice on most large sites seems to not send 404s to a trailing slash if the page doesn't exist since this is an unnecessary redirect and may confuse the user. That's another reason I'd like to see this behavior phased out.

stephenmcd:
Closing in light of Alex's comments: https://github.com/stephenmcd/mezzanine/pull/1437#issuecomment-163818861

PR #1444, opened 2015-10-26
Check for FORMS_USE_HTML5 on the admin instead of the model

### LINKED ISSUE #1399: FORMS_USE_HTML5 causes a migration
Setting `FORMS_USE_HTML5 = True` will trigger a migration, because the `editable` attribute for the placeholder field in [models.py](https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/forms/models.py#L68) depends on it.

It looks like the purpose behind the conditional `editable` is to easily show/hide the placeholder in the `Field` inline admin. Is that correct, or am I missing something?

jerivas:
Fixes #1399.

I decided to go with the the `exclude` attribute since this is an inline model admin. It's let's verbose too and if new fields are added to the Field model, they will just show up without having to explicitly include them.

I tested with modeltranslation enabled and it works fine.

### DIFF
--- a/mezzanine/forms/admin.py
+++ b/mezzanine/forms/admin.py
@@ -35,6 +35,10 @@ form_fieldsets = list(form_fieldsets)
 form_fieldsets.insert(1, (_("Email"), {"fields": ("send_email", "email_from",
     "email_copies", "email_subject", "email_message")}))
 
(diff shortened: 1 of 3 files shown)


stephenmcd:
Nice one, thanks Ed.

PR #1463, opened 2015-11-21
Document errors_for and update fields_for docs.

ryneeverett:
- The `fields_for` template path and type are out of date.
- Document `errors_for` since it adds a lot of value to `fields_for`.

Not sure how we feel about adding additional template tags to the documented API but I thought I'd throw it out there.


stephenmcd:
Awesome, thanks Ryne!

PR #1468, opened 2015-11-28
1148 improve api docs

chmelevskij:
Added inner-references to:
* Content Architecture
* Model Costumization
* Admin Costumization


stephenmcd:
Thanks a lot for this!

I'm not sure how you've managed to have this reference so many already existing commits, but you'll need to sort that out and re-submit it. That will also give you the chance to squash your documentation updates into a single commit with one clean commit message.

PR #1469, opened 2015-11-30
Add references to api-docs

chmelevskij:
This will add links to documentation  generated from the source. This only adds links to the things which are included in the configuration.rst file.


stephenmcd:
Great stuff, thanks a lot Tomas.

PR #1473, opened 2015-12-01
Don't strip HTML in TinyMCE.

ryneeverett:
Because filtering is handled by bleach.
Reimplement 0f6ab7c.

To quote Paul Whipp on the mailing list:
> If I say 'no filtering' I mean don't mess with my html dammit :)


stephenmcd:
Thanks Ryne - that looks familiar, I think we lost that during the TinyMCE upgrade.

Issue #1474, opened 2015-12-01
Cannot access comments without rendering the comments thread template


JDougherty:
Currently, Mezzanine has the [comment_thread](https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/generic/templatetags/comment_tags.py#L37) templatetag which loads the comment objects for a given parent object and renders a template for those comments. It has some nice logic to get all the comments for staff users, and only visible comments otherwise. However, as an inclusion tag, when used it renders the included template in the location it was called. This means you cannot access the comment objects for a given parent without rendering a template. 

I feel it is desirable to be able to access the objects directly, without being forced to render a template. I propose a new assignment tag that will access the comment objects and place them in the context for use in the template. The `comment_thread` tag already checks for `all_comments` in the context, so using an assignment tag can be a sort of pre-load of the objects. If `all_comments` isn't in the context already, the `comment_thread` tag can use the new assignment tag directly. 

I have a fix in progress, I will add a pull request shortly.

JDougherty:
As you can see in the diff, all I've really done is move the logic that loads the comment objects from `comment_thread` to a new assignment tag `comment_thread_objects`. This allows a developer to access the objects directly in a template like so:

    {% comment_thread_objects blog_post as all_comments %}

If used as above, with `as all_comments`, then the `comment_thread` tag does not need to perform the lookup again as it's already in the context. If the objects have not been loaded into the context as `all_comments`, then the `comment_thread` tag will load them as normal.

stephenmcd:
I've left feedback on the PR - BTW no need to open a separate issue.

PR #1475, opened 2015-12-01
Created new assignment tag comment_thread_objects

JDougherty:
For issue #1474


stephenmcd:
Hi, this looks good but we probably don't need it in Mezzanine by default.

If you really want to make it available to the community you could release it as a plugin.

Thanks!

JDougherty:
Thanks for the feedback @stephenmcd. Is there documentation on how to make a plugin for the community? Thanks!

PR #1477, opened 2015-12-02
Recommend using includes/footer_scripts.

ryneeverett:
This is how it's done in the default templates but docs-readers probably
aren't going to know that they're already calling `editable_loader` via
an include.


stephenmcd:
Thanks Ryne.

PR #1478, opened 2015-12-02
Tighten up edge cases in password reset form

dsanders11:
There's three different cases being tightened up here.
1. Respect `ACCOUNTS_NO_USERNAME` and don't allow resetting by username if it's set to `True`.
2. If `username` evaluates to `False`, immediately return a `ValidationError` without a DB query. Some databases (such as MySQL) will take any number of blank spaces and return all rows with empty values. If accounts exist with no email address (created via admin) then a password reset request on ' ' will return all of those accounts from the DB and error due to too many results.
3. Handle legitimate multiple results. While username is unique, emails are not constrained to unique values. Manual entry in admin may lead to multiple accounts with the same email. Don't let the exception bubble up in this case and instead use a `ValidationError` to provide a better user experience.

### DIFF
--- a/mezzanine/accounts/forms.py
+++ b/mezzanine/accounts/forms.py
@@ -249,12 +249,22 @@ class PasswordResetForm(Html5Mixin, forms.Form):
 
     def clean(self):
         username = self.cleaned_data.get("username")
-        username_or_email = Q(username=username) | Q(email=username)
+        email_or_username = Q(email=username)
+
+        if not settings.ACCOUNTS_NO_USERNAME:
+            email_or_username |= Q(username=username)
+
         try:
-            user = User.objects.get(username_or_email, is_active=True)
+            if not username:
+                raise User.DoesNotExist()
+
+            user = User.objects.get(email_or_username, is_active=True)
         except User.DoesNotExist:
             raise forms.ValidationError(
                              ugettext("Invalid username/email"))
+        except User.MultipleObjectsReturned:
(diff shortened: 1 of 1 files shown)


molokov:
Another thing I've noticed with some of my users is that for whatever reason, email address is case sensitive - so I've had people saying they can't reset their password by email address simply because they used an uppercase letter in their address when making an account, but not when trying to reset the password. Emails should be case insensitive.

dsanders11:
@molokov, that's a good point, you're right, it looks like emails are indeed case sensitive on reset and what not. That's probably a separate pull request since the merits would need to be debated (technically emails have to be case sensitive but in the real world they rarely are) and it would need to be enforced throughout Mezzanine (user signup, updating email, etc).

AlexHill:
I think it's fine in practise to consider emails completely case-insensitive for the purposes of uniqueness.

A good way to go here is to do what dabapps/django-email-as-username does IIRC – generate a username from a hash of the lowercased email address. That way you preserve the casing that the user entered, and still ensure uniqueness and case-insensitive lookups.

See here: https://github.com/dabapps/django-email-as-username/blob/master/emailusernames/utils.py#L15

stephenmcd:
I'm not 100% sure this is right - I think `ACCOUNTS_NO_USERNAME` is meant to remove any support for usernames, but you may still want a username while being able to login via email.

Thoughts?

PR #1479, opened 2015-12-04
Added setting for no override the runserver command

AliLozano:
Runserver command of mezzanine doesn't should override the runserver command if I don't want, If I have already a media configuration, I don't need a custom runserver.
I added a setting for no override the runserver command.


stephenmcd:
Sorry I forgot to give feedback - I just don't think this warrants a setting to be honest, if you really want you can hack it out in your own project.

Issue #1483, opened 2015-12-07
ValueError while import blog posts from Blogger 


dablak:
Hi, when I run the Blogger importer: 

```
$ python manage.py import_blogger --mezzanine-user=.. --blogger-id=XXX
```

A ValueError is raised:

```python
Traceback (most recent call last):
  File "manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/.../django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/.../django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/.../django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/.../django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/.../mezzanine/blog/management/base.py", line 168, in handle
    self.handle_import(options)
  File "/.../mezzanine/blog/management/commands/import_blogger.py", line 59, in handle_import
    "%Y-%m-%dT%H:%M:%S.%f") - timedelta(seconds=timezone)
  File "/.../python2.7/_strptime.py", line 325, in _strptime
    (data_string, format))
ValueError: time data '2015-11-26T16:21:0' does not match format '%Y-%m-%dT%H:%M:%S.%f'
```

A possible way of fixing this is to change mezzanine/blog/management/commands/import_blogger.py

```python
try:
    published_date = datetime.strptime(entry.published.text[:-6],
            "%Y-%m-%dT%H:%M:%S.%f") - timedelta(seconds=timezone)
except ValueError:
    published_date = datetime.strptime(entry.published.text[:-6],
            "%Y-%m-%dT%H:%M:%S") - timedelta(seconds=timezone)
```

and

```python
try:
    comment_date = datetime.strptime(comment.published.text[:-6],
        "%Y-%m-%dT%H:%M:%S.%f") - timedelta(seconds=timezone)
except ValueError:
    comment_date = datetime.strptime(comment.published.text[:-6],
        "%Y-%m-%dT%H:%M:%S") - timedelta(seconds=timezone)
```

PR #1484, opened 2015-12-07
Fix ValueError when importing blog posts from Blogger

dablak:
Fixed ValueError exception when parsing dates of imported Blogger posts. Issue #1483


stephenmcd:
Since you've repeated the same code twice, how about putting it into a function and calling that from both spots?

dablak:
Sure, good idea. There you go!

PS. It took me a few commits to have all the right line lengths and unnecessary spaces removed from the code. Very picky this mr. Travis hehehe

stephenmcd:
Great, thanks. Can you also squash into a single commit with a proper commit message? I can merge it then.

dablak:
There you go

stephenmcd:
Absolutely perfect, thanks!

PR #1489, opened 2015-12-10
adding missing french translations in accounts

flo-dhalluin:
Hello, 

While playing around, I noticed some french messages where missing from acounts (like when requiring manual activation). 

I didn't compile the .mo, let me know if I need to compile it.


stephenmcd:
You'll see we include the compiled .mo files - can you update that too? Thanks

flo-dhalluin:
There you are

stephenmcd:
Thanks a lot.

PR #1493, opened 2015-12-11
Add references to configuration.rst

chmelevskij:
Any mezzanine defined setting are referencing to configuration.rst


stephenmcd:
Thank you! Couple of points:

Shouldn't `from string import ascii_letters` be `from string import ascii_letters as letters` ?

Is it necessary for the settings labels to be different from the settings? Eg do we have to use `SETTING-LABEL` instead of just `SETTING`? It would be much easier to read as text then. If it has to be different maybe it should be `REF-SETTING` which matches more closely what's happening.

stephenmcd:
Oh I just saw the previous discussion where someone suggested the `-LABEL` suffix. I think adding `ref` is good but again if it's not necessary, we should just use ```:ref:`SETTING` ``` instead of ```:ref:`SETTING-LABEL` ```

stephenmcd:
And once everything's done please squash into a single commit.

Thanks again!

chmelevskij:
I've done the ```from string import ascii_letters as letters``` in the last commit. Will do other suggestions when by the end of the week when.

stephenmcd:
Awesome, no rush :-)

PR #1495, opened 2015-12-13
Move SingletonAdmin to utils. Resolve #1480.

### LINKED ISSUE #1480: SingletonAdmin can't be imported before models are loaded.
When I import `SingletonAdmin` in a file that gets sourced at startup (models.py) I get the following exception:
```py
...
  File ".../cartridge-example/app_name/models.py", line 3, in <module>
    from mezzanine.core.admin import SingletonAdmin
  File ".../python3.5/site-packages/mezzanine/core/admin.py", line 54, in <module>
    User = get_user_model()
  File ".../python3.5/site-packages/django/contrib/auth/__init__.py", line 150, in get_user_model
    return django_apps.get_model(settings.AUTH_USER_MODEL)
  File ".../python3.5/site-packages/django/apps/registry.py", line 199, in get_model
    self.check_models_ready()
  File ".../python3.5/site-packages/django/apps/registry.py", line 131, in check_models_ready
    raise AppRegistryNotReady("Models aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.
```

It can easily be worked around by doing the registration in a model that isn't sourced at startup such as `admin.py`, but it seems less than ideal for a documented API to have that limitation.

This could probably be fixed with a try/except block in mezzanine/core/admin, but I wonder if it wouldn't be better to move `SingletonAdmin` to a new file or even a new app (contrib?) since it isn't used at all internally. I'd be willing to try implementing a solution if one were agreed upon.

ryneeverett:
Unfortunately I'm not aware of any reasonable way to raise a `DeprecationWarning` on import, but I would suppose importing `SingletonAdmin` in `core.admin` will suffice until it we can remove it in a major version.

### DIFF
--- a/docs/admin-customization.rst
+++ b/docs/admin-customization.rst
@@ -220,7 +220,7 @@ from your custom widget, by doing the following:
 Singleton Admin
 ===============
 
-The admin class :class:`mezzanine.core.admin.SingletonAdmin` is a utility
+The admin class :class:`mezzanine.utils.admin.SingletonAdmin` is a utility
 that can be used to create an admin interface for managing the case
 where only a single instance of a model should exist. Some cases
 include a single page site, where only a few fixed blocks of text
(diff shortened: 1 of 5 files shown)


stephenmcd:
Looks perfect, thanks Ryne.

PR #1500, opened 2015-12-21
Add classes to app_list in admin

julianandrews:
This provides a hook for model specific styling, and mirrors what Django
itself does. See:

https://github.com/django/django/blob/5399ccc0f4257676981ef7937ea84be36f7058a6/django/contrib/admin/templates/admin/index.html#L23


stephenmcd:
Nice one, thank you.

PR #1501, opened 2015-12-22
Catch error if user sitepermissions don't exist.

julianandrews:
For the most part, if not using `SitePermissionMiddleware` Mezzanine
falls back on the `is_staff` attribute seamlessly. But since
d5d21ba527bd4 it's possible to have users without sitepermissions. This
breaks the admin dropdown, but not much else. This fix should allow
single site projects to continue to leave `SitePermissionMiddleware`
uninstalled.

I think this is a reasonable way to fix this issue (and is a reasonable issue to fix) but I'll admit I don't have the entirety of the `SitePermission` setup clear in my head, so I'm very open to suggestions for better approaches. If at all possible it would be really helpful to be able to preserve the `is_staff` behavior for single site projects.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -12,6 +12,7 @@ from django.apps import apps
 from django.contrib import admin
 from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.contrib.sites.models import Site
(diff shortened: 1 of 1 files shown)


stephenmcd:
Awesome, this actually caught me recently and I hadn't looked at it yet. Thanks a lot!

PR #1506, opened 2016-01-05
Fix app list in the Admin

acatton:
Using mezzanine breaks the admin in multiple ways:

* On the admin dashboard, there's no link to the app specific listing. (The header of the section) But you can access the app specific listing by clicking on the breadcrumb.
* When listing an app, the full list is getting displayed. (Because the default `app_index.html` inherit from the default templates which displays the dashboard)

This pull request fixes these two bugs.

I also removed the `{%  trans %}` tag on the admin header because you can't really translate variables, and names are already translated.

### DIFF
--- a/mezzanine/core/templates/admin/app_index.html
+++ b/mezzanine/core/templates/admin/app_index.html
@@ -1,0 +1,22 @@
+{% extends "admin/index.html" %}
+{% load i18n %}
+
+{% block bodyclass %}{{ block.super }} app-{{ app_label }}{% endblock %}
+
+{% if not is_popup %}
+{% block breadcrumbs %}
+<div class="breadcrumbs">
+<a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
+&rsaquo;
+{% for app in app_list %}
+{{ app.name }}
+{% endfor %}
+</div>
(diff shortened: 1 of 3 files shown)


stephenmcd:
These were all intentionally moved - since we always have the left-hand navigation available, they're of no use.

Thanks all the same.

gavinwahl:
Even though you have links in the left-hand nav, what's the harm in having the links from Django admin too?

The breadcrumb also links to the app index page.

stephenmcd:
> Even though you have links in the left-hand nav, what's the harm in having the links from Django admin too?

It's redundant.

> 
> The breadcrumb also links to the app index page.
> 

These are meant to be removed (also redundant). If they're showing anywhere, that's a bug.

> I'm trying to use Mezzanine without Grappelli (something that Mezzanine seems to attempt to support with GRAPPELLI_INSTALLED), but the overridden templates Mezzanine provides cause this breakage.

That was once a goal but it's not something I've personally kept up with for years now - over time we've reduced the number of "variations" in environments in order to reduce the cost of maintenance, the number of supported Django versions being the main example, but this is certainly another.

But hey, if you need these kinds of fixes, by all means do something similar to this PR within your project.

PR #1510, opened 2016-01-10
Project template fix RemovedInDjango110Warning's

ryneeverett:


stephenmcd:
Thanks Ryne. I tweaked the commented-out patterns a bit in urls.py, I thought it better to put a note in the comments about importing the views rather than having commented out imports at the top that didn't obviously related to the commented out patterns way down below.

Issue #1514, opened 2016-01-20
django fails replacing FK "sites.Site" string with actual class


gradel:
After facing this exception in my project could reproduce it on a plain testing install:

```
Internal Server Error: /sitemap.xml
Traceback (most recent call last):
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/db/models/base.py", line 401, in __init__
    rel_obj = kwargs.pop(field.name)
KeyError: 'site'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/db/models/base.py", line 406, in __init__
    val = kwargs.pop(field.attname)
KeyError: 'site_id'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/core/handlers/base.py", line 149, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/contrib/sitemaps/views.py", line 16, in inner
    response = func(request, *args, **kwargs)
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/contrib/sitemaps/views.py", line 67, in sitemap
    protocol=req_protocol))
  File "/home/gerald/tests/mezzanine/mezzanine/mezzanine/core/sitemaps.py", line 38, in get_urls
    return super(DisplayableSitemap, self).get_urls(**kwargs)
  File "/home/gerald/tests/mezzanine/env/lib/python3.4/site-packages/Django-1.9.1-py3.4.egg/django/contrib/sitemaps/__init__.py", line 104, in get_urls
    urls = self._urls(page, protocol, domain)
[...]

PR #1516, opened 2016-01-26
Test and fix for displayable_links_js

viaregio:
I think this is also related to issue #1514.


stephenmcd:
Don't think this fix was quite right, I just committed what I think was more correct.

Thanks!

molokov:
A related issue - the above commit does indeed fix the exception we were getting, and displayable_links now works when editing/inserting a link in TinyMCE from the 'front-end' editing.

However, I'm not getting the list of links when inserting/editing a link when editing a page/blog post through the Admin side of things - no exception thrown, but the list of links isn't appearing. 

Do you want me to raise this as a separate issue?

jjmiranda:
@stephenmcd any more correct solution for isinstance() exception in  /displayable_links.js to commit in your master?
Thank you very much for your time on this.

michela:
Popup link editor still broken in latest Mezzanine commit (43036df0ff8be70e44d5116ea467f01432867108). Noticed that the test wasn't merged from that pull  - this test hasn't been merged

+    def test_displayable_links(self):
+        self.client.login(username=self._username, password=self._password)
+        links_url = reverse('displayable_links_js')
+        response = self.client.get(links_url)
+        self.assertEqual(response.status_code, 200)

PR #1517, opened 2016-01-28
Fixes #1449 -- duplicate profile fields in admin

### LINKED ISSUE #1449: Duplicate form fields in admin for user profiles
As discussed here:

https://groups.google.com/forum/#!topic/mezzanine-users/3QmiqfNZjUM

AlexHill:
Same as the previous fix in #1450, but with ``min_num = 1`` added to ``ProfileInline``.

### DIFF
--- a/mezzanine/accounts/admin.py
+++ b/mezzanine/accounts/admin.py
@@ -51,6 +51,7 @@ try:
         can_delete = False
         template = "admin/profile_inline.html"
         extra = 0
+        min_num = 1
     UserProfileAdmin.inlines += (ProfileInline,)
 except ProfileNotConfigured:
     pass
--- a/mezzanine/accounts/admin.py
+++ b/mezzanine/accounts/admin.py
@@ -51,7 +51,13 @@ try:
         can_delete = False
         template = "admin/profile_inline.html"
         extra = 0
-        min_num = 1
+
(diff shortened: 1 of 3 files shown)


AlexHill:
I've moved the `empty-form` CSS rule to grappelli-safe, where I think it belongs. The CSS rule is already in vanilla Django, and the reason we don't see it is because of grappelli's overrides, so that's where the fix should live IMO.

See stephenmcd/grappelli-safe#65

stephenmcd:
Testing this and it seems to prevent adding new users - when I try to, I'm shown the add user form again but without any error messages. 

I've a profile model with a required field - my rough guess is that the form for adding users thinks that profile field should have a value, even though it shouldn't appear on the add user form.

AlexHill:
Damn, I'll have another look. Adding worked ok on my setup but with no
required profile fields.

So it's the initial username/password form you can't get past?

stephenmcd:
Yep

AlexHill:
Ok, well that's definitely what's happening.

Fortunately `InlineModelAdmin` provides a `get_min_num()` method which we can override, returning 0 when creating and 1 when updating. That should sort it out, can you give it a go Steve?

stephenmcd:
Works! Thanks as always for diving deeper than me.

PR #1519, opened 2016-01-28
Removed nesting of Context object in editable_loader

spookylukey:
The additional wrapping of the input `Context` object inside another `Context` object is unnecessary, and causes problems.

For example, `flatten` no longer works:

    >>> from django.template.context import Context
    >>> Context(Context()).flatten()
    ValueError: dictionary update sequence element #0 has length 3; 2 is required

In particular, this causes an incompatibility with django-compressor 1.6+

Producing a full test case of how it causes problems would be a lot of work - any chance this could be accepted without that? There aren't many existing tests for this bit of code - or any, AFAICS.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -422,7 +422,7 @@ def editable_loader(context):
             context["editable_obj"]
         except KeyError:
             context["editable_obj"] = context.get("page", None)
-        context["toolbar"] = t.render(Context(context))
+        context["toolbar"] = t.render(context)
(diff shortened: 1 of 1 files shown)


spookylukey:
Actually, the problem is bigger than this.

There is a problem with many usages of `inclusion_tag`. Both Django's version and Mezzanine's version of this function wrap the result of the inclusion tag function in a Context. This is a problem if the function itself returns a Context object.

The way that inclusion_tags are supposed to work is that they should return dictionaries:

https://docs.djangoproject.com/en/1.9/howto/custom-template-tags/#inclusion-tags

I'll work on a proper patch.

stephenmcd:
Closing in favour of #1520

PR #1520, opened 2016-01-28
Removed nesting of Context objects

spookylukey:
This supersedes https://github.com/stephenmcd/mezzanine/pull/1519 

Inclusion tags should:

* Return dictionaries, not Context objects
* Explicitly pass all needed data to their templates
* Not modify passed in Context object ideally

No code should wrap Context objects in further Context objects, as this
causes errors in some contexts.

----

Details

The additional wrapping of the input Context object inside another Context object is unnecessary, and causes problems.

For example, flatten no longer works:

    >>> from django.template.context import Context
    >>> Context(Context()).flatten()
    ValueError: dictionary update sequence element #0 has length 3; 2 is required

In particular, this causes an incompatibility with django-compressor 1.6+, and possibly various other projects, especially if you need to override the templates Mezzanine provides for these inclusion tags.

In this PR, I have tried to make sure context dictionaries with everything necessary for the templates are created, and have then manually checked that things like the editable loader etc. are not broken. However, the lack of automated test coverage for this code means that more manual testing might be needed.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -101,7 +101,7 @@ def fields_for(context, form, template="includes/form_fields.html"):
     Renders fields for a form with an optional template choice.
     """
     context["form_for_fields"] = form
-    return get_template(template).render(Context(context))
+    return get_template(template).render(context)
 
 
 @register.inclusion_tag("includes/form_errors.html", takes_context=True)
@@ -109,8 +109,7 @@ def errors_for(context, form):
     """
     Renders an alert if the form has any errors.
     """
-    context["form"] = form
-    return context
+    return {"form": form}
 
 
 @register.filter
@@ -244,6 +243,9 @@ def search_form(context, search_model_names=None):
     string ``all`` can also be used, in which case the models defined
     by the ``SEARCH_MODEL_CHOICES`` setting will be used.
     """
+    c = {
+        "request": context["request"],
+    }
     if not search_model_names or not settings.SEARCH_MODEL_CHOICES:
         search_model_names = []
     elif search_model_names == "all":
@@ -259,8 +261,8 @@ def search_form(context, search_model_names=None):
         else:
             verbose_name = model._meta.verbose_name_plural.capitalize()
             search_model_choices.append((verbose_name, model_name))
-    context["search_model_choices"] = sorted(search_model_choices)
-    return context
+    c["search_model_choices"] = sorted(search_model_choices)
+    return c
 
 
 @register.simple_tag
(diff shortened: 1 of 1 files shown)


mezzanine/core/templatetags/mezzanine_tags.py
```diff
@@ -109,8 +109,7 @@ def errors_for(context, form):
     """
     Renders an alert if the form has any errors.
     """
-    context["form"] = form
-    return context
+    return {"form": form}
```
henri-hulski: Are you not loosing here the context which is parsed into the function?

spookylukey: Yes, but that is how inclusion tags are supposed to work - https://docs.djangoproject.com/en/1.9/howto/custom-template-tags/#inclusion-tags - you are supposed to return a simple dictionary containing only the data needed, not the whole context.

The only value used by the `form_errors.html` template is `form`, so behaviour is preserved.

The alternatives:
1) what happened before - that you actually mutate the context that is passed in, which is a bad idea for the sake of the calling context, because it potentially has to deal with the side effects of the inclusion tag, and the template gets a nested Context object.
2) copy all the data from the passed in Context obejct into a new Context object. This seems like a bad idea when you only need one value from it.

stephenmcd:
Hi Luke, thanks for looking at this.

Just to clarify, are there two separate changes happening here?

1) Removal of nested context objects
2) Changing return type of some template tags from contexts to dicts

Is it only 1) that related to breaking `compressor`? If so I think 2) needs a great deal more discussion and should be separated from 1). 

For more context (literally, not technically) we actually have a bunch of template tags (specifically `page_menu`) that make heavy use of modifying the global template context and returning context objects. If we're changing some template tags, shouldn't we change them all? It feels like some consistency should be here. I also think that even though this goes against the docs you linked to, exposing all context vars to template includes is a good feature - something that some users may already be dependent on in their own templates that override ours. Doesn't seem like this is worth potentially breaking yet, if this change isn't strictly necessary.

spookylukey:
Yes, there a few sets of related changes - I thought it worthwhile addressing all instances of "Context inside Context" that I could see in that module, I didn't realise there were lots more instances. Returning Context objects from `inclusion_tag` functions results in nested Context because of the implementation of `inclusion_tag`, both Mezzanine's version and Django's:

https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/template/__init__.py#L121
https://github.com/django/django/blob/1.9/django/template/library.py#L236

For my own needs, it looks like the fixes for `editable_loader` are sufficient, particularly that it returns a dictionary and not a Context. We are overriding `editable_loader.html` to add a Javascript resource, and using a `compress` tag from django-compressor. With django-compressor 1.6 or later, this breaks because of the nested Context object.

I'm bringing the bugfix here because it seems that it is Mezzanine that is at fault - putting a Context instance inside a Context instance is just not properly supported by Django, and not the intended usage, although it works a lot of the time. I suspect this is only going to get worse with Django 1.8 and later, which are more fussy about using dicts not Context in some places - for example:

https://github.com/django/django/blob/1.8/django/template/backends/django.py#L66

You're right that there is the compatibility issue with other people who may have overridden the templates and are expecting more data. The other side is that exposing all the parent data into the included template is asking for trouble in terms of compatibility - without documentation, you have no idea what context should be there and might be depended on, so you are never allowed to remove anything from the context that you pass in to any views. From a consumer's point of view, it isn't even documented that these template tags can be customized by providing your own template, like we are doing, so technically you wouldn't be breaking a backwards compatibility promise.

Moving forward, I think it would be best to define for each template tag what template is used by default and what context is available clearly, and to start it with only what is necessary - potentially adding `request` in to everything, because that is often used. An alternative is to regard the whole implementation of provided template tags to be a black box. For our own purposes, we would then provide our own implementation of the tag function as well as our own template for it, in which case no changes to Mezzanine are necessary at all. As long as we know what the policy is, we can plan accordingly.

I'm afraid I'm not likely to be able to do much more on this in terms of patches, especially as there don't seem to be tests for these template tags to use as a starting point.

Thanks for your time!

stephenmcd:
Thanks Luke, that's a clear argument. I'll merge the changes.

Issue #1525, opened 2016-02-02
Remove SSL3 support from nginx.conf.template


douglaskastle:
I was doing a security review of a base install website using this tool:

https://www.ssllabs.com/ssltest/analyze.html

It flagged that SSL3 was a major weakness due to the POODLE attack:

https://blog.qualys.com/ssllabs/2014/10/15/ssl-3-is-dead-killed-by-the-poodle-attack

And recommends disabling SSL3 in future installs.  This can be acheieved by adding this line to nginx.conf (and also the template):

    %(ssl_disabled)s ssl_protocols TLSv1.2 TLSv1.1 TLSv1;

stephenmcd:
Could you create a PR with the change? Thanks for your help.

PR #1526, opened 2016-02-02
Removing support for SSL3

douglaskastle:
SSL3 is susceptible to the POODLE attack:

http://disablessl3.com/

Added line to the nginx.conf.template to remove this support from the default project.

This is captured in bug #1525


stephenmcd:
Thanks Douglas

Issue #1523, opened 2016-02-01
Remove syncdb from default fabfile


douglaskastle:
It looks like Django 1.9 won't support syncdb as a command any more (though it supposed to have had a depreciation warning since 1.7). A default project deploy with fabric falls over.

I think it is sufficient to delete line 638 from /mezzanine/project_template/fabfile.py the next command, migrate, is doing the correct job now.

douglaskastle:
This fix has been merged, closing.

PR #1527, opened 2016-02-02
Removed syncdb from deploy as it is no longer supported in Django1.9 

douglaskastle:
It has been depricated since 1.7, migrate is used instead.

This is captured in bug #1523


stephenmcd:
Thanks!

PR #1529, opened 2016-02-02
import blogml

emattiza:
[BlogML Schema Definition](https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/blogml/BlogML.xsd)

---

I personally used some of this to handle some imports, but the particular cms used BlogML schema in some quirky ways, so modifications were made to be more in-line with the schema def and not some custom cms baloney.

Currently no comment handling implemented. If i had sample data with comments, i would like to test it out and try some things. Structure is there.

R0914 for too many locals in handle_import as well (17/15). Unsure of style guidelines, but default pylint settings were used.

### DIFF
--- a/mezzanine/blog/management/commands/import_blogml.py
+++ b/mezzanine/blog/management/commands/import_blogml.py
@@ -5,8 +5,6 @@ Python module to implement xml parse and import of blogml blog post data
 """
 import xml.etree.ElementTree as ET
 
-from dateutil.parser import parse
-
 from mezzanine.blog.management.base import BaseImporterCommand
 
 
@@ -31,6 +29,12 @@ class Command(BaseImporterCommand):
         - options is an optparse object with one relevant param
          * xmlfname is for path to file
         """
+
+        try:
+            from dateutil.parser import parse
+        except ImportError:
+            raise ImportError("parse not found in dateutil; python-dateutil")
+
         xmlfname = options.get("xmlfilename")
         # parsing xml tree and populating variables for post addition
         tree = ET.parse(xmlfname)
@@ -91,7 +95,8 @@ class Command(BaseImporterCommand):
                 (categories_found[x] for x in post_category_refs))
             found["tags"] = [x.attrib["ref"] for x in
                              post.findall("blogml:tags/blogml:tag",
-                                          namespaces=ns)]
+                                          namespaces=ns)
+                             ]
             post_entered = self.add_post(title=found["title"],
                                          content=found["content"],
                                          old_url=found["post-url"],
--- a/mezzanine/blog/management/commands/import_blogml.py
+++ b/mezzanine/blog/management/commands/import_blogml.py
@@ -16,12 +16,12 @@ from mezzanine.blog.management.base import BaseImporterCommand
 class Command(BaseImporterCommand):
     """
     This class extends django management and mezzanine custom blog import
-    commands to allow for import from blogml styled blogs
-    Should supply user to upload as and xmlfilename
+    commands to allow for import from BlogMl styled blogs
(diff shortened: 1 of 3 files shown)


stephenmcd:
Looks great, couple nitpicks:

- Please use double quotes for docstrings for consistency
- Spaces between import groups (stdlib, thirdparty (eg django), internal (eg mezzanine))
- Why is the timezone option mandatory, shouldn't it default to whatever's configured via Django?

stephenmcd:
Also docs for blog importing need updating, each command has a brief section: https://github.com/stephenmcd/mezzanine/blob/master/docs/blog-importing.rst

emattiza:
My reasoning was that perhaps there may be a situation where a timezone would be inherently different in an xml dump (say UTC, giving the user the option to properly select the timezone format used in the dump). 

Localizing to django-provided settings with the assumption that the two match is a good enough bet.

stephenmcd:
Sure, I'm not questioning having the option, it makes sense - but it should be optional.

mezzanine/blog/management/commands/import_blogml.py
```diff
@@ -0,0 +1,74 @@
...
+        make_option("-x", "--blogxmlfname", dest="xmlfilename",
+                    help="xml file to import blog from BoxAlly"),
+        make_option("-z", "--timezone", dest="tzinput",
+                    default=timezone.get_current_timezone_name())
+    )
+
+    def handle_import(self, options):
+        """
+        Gets posts from provided xml dump
+        TODO: Handle comment importing
```
stephenmcd: Does blogml support comments? If so we should support that before merging this.

emattiza:
It's got a pretty large support base. See the xsd for more details. I'll see about expanding the scope and cleaning it up more.

setup.py
```diff
@@ -60,6 +60,7 @@
             "requests >= 2.1.0",
             "requests-oauthlib >= 0.4",
             "future >= 0.9.0",
+            "python-dateutil",
```
ryneeverett: I don't think we should be adding hard requirements for blog importing, which wont be needed most of the time. It could be an extra like `extras_require={'blogimport': ['python-dateutil']},`

mezzanine/blog/management/commands/import_blogml.py
```diff
@@ -0,0 +1,127 @@
...
+                             post.findall("blogml:tags/blogml:tag",
+                                          namespaces=ns)]
+            post_entered = self.add_post(title=found["title"],
+                                         content=found["content"],
+                                         old_url=found["post-url"],
+                                         pub_date=found["date-created"],
+                                         categories=found["categories"])
+            found.clear()
+            comments_found = post.findall(search_paths["find_comments"])
+            cmmnt = {}
```
stephenmcd: Let's use proper variable names. I see there's already a `comment` variable. How about changing them to `comment_dict` and`coment_obj`?

mezzanine/blog/management/commands/import_blogml.py
```diff
@@ -0,0 +1,127 @@
+"""
+Python module to implement xml parse and import of blogml blog post data
+
+ * Has dependency of python-dateutil
+"""
+import xml.etree.ElementTree as ET
```
stephenmcd: Don't think this needs an alias.

setup.py
stephenmcd: Agreed.

@emattiza take a look at some of the other blog importers, you'll see they just wrap the import of the optional dependency with `try/catch ImportError` and if it's missing they raise an appropriate error message.

stephenmcd:
Looks good! Thanks a lot for all your work, and for addressing all the feedback.

PR #1538, opened 2016-02-13
Fixes #1438: allow multiple comment threads on the same page

### LINKED ISSUE #1438: Returns same comment form for different model when requesting from same template.
This error occurs when we  request for comments for more than one models from same template.
Use case :
Let's initially we request for comment for model Question in the same context again request for comment for Answer model then it returns same form since try catch statement bypasses update of form.

Code from  /mezzanine/generic/templatetags/comment_tags.py

```python
def comments_for(context, obj):
...
26    try:
27        context["posted_comment_form"]
28    except KeyError:
29        context["posted_comment_form"] = form
30    context["unposted_comment_form"] = form
31    ...
```
Error at line number 26.
Small fixed for this.
Replace try catch block with below code.
```python
if context.get("posted_comment_form") is None or \
                    context.get("posted_comment_form") != form:
        context["posted_comment_form"] = form
        context["unposted_comment_form"] = form 
```

sonus21:
This will fix bug #1438 
Changes:
```python
def comments_for(context, obj):
...
26    try:
27        context["posted_comment_form"]
28    except KeyError:
29        context["posted_comment_form"] = form
30    context["unposted_comment_form"] = form
31    ...
```
to 
```python
if context.get("posted_comment_form") is None or \
                    context.get("posted_comment_form") != form:
        context["posted_comment_form"] = form
        context["unposted_comment_form"] = form 
```

As I mentioned here #1438 it returns same comment form. Now we will check for two conditions whether it's a new comment form or a comment form has to be updates.

### DIFF
--- a/mezzanine/generic/templatetags/comment_tags.py
+++ b/mezzanine/generic/templatetags/comment_tags.py
@@ -23,11 +23,11 @@ def comments_for(context, obj):
     """
     form_class = import_dotted_path(settings.COMMENT_FORM_CLASS)
     form = form_class(context["request"], obj)
-    try:
-        context["posted_comment_form"]
-    except KeyError:
+    if context.get("posted_comment_form") is None or \
+                    context.get("posted_comment_form") != form:
         context["posted_comment_form"] = form
-    context["unposted_comment_form"] = form
+        context["unposted_comment_form"] = form 
+
     context["comment_url"] = reverse("comment")
     context["object_for_comments"] = obj
     return context


stephenmcd:
Can you explain the change?

sonus21:
@stephenmcd See once I have updated commit details.

mezzanine/generic/templatetags/comment_tags.py
```diff
@@ -23,11 +23,11 @@ def comments_for(context, obj):
     """
     form_class = import_dotted_path(settings.COMMENT_FORM_CLASS)
     form = form_class(context["request"], obj)
-    try:
-        context["posted_comment_form"]
-    except KeyError:
+    if context.get("posted_comment_form") is None or \
+                    context.get("posted_comment_form") != form:
```
AlexHill: I'm pretty sure this will always evaluate `True`. That is, the form we just instantiated locally will never be equal to the form instantiated in the `comment` view and passed into the context.

A more useful check might be to compare `form.target_object`?

```python

context_form = context.get("posted_comment_form")

return {
    'posted_comment_form': context_form if context_form.target_object == obj else form,
    'unposted_comment_form': form,
    'comment_url': reverse("comment"),
    'object_for_comments': obj,
    'request': context.get('request'),
}
```

I'm returning a new context here to avoid introducing context variables into the global context that are only relevant to `generic/includes/comments.html`.

sonus21: I have tested this and you can see on my site http://www.job360.io/question-answer/cse/how-to-add-two-large-numbers, it's  working perfectly.

AlexHill:
Fixed in a0a205e3121835a4f423eac7dc4d581eb6f0df5d

PR #1540, opened 2016-02-20
Accept either CDATA or text nodes for WordPress comments.

sjuxax:
All text output is CDATA in an export from WordPress 4.4.2. Had to update this script to get import to proceed.

It is hoped that accepting either text or CDATA as node type will be compatible with exports from all versions of WordPress.


stephenmcd:
Thanks Jeff

PR #1547, opened 2016-02-27
Add Mime Type determination to static_proxy

zandeez:
I was having issues with uploadify because asset_proxy was setting Content-type to application/octet-stream, so I've added code to use the mimetypes library to detemine the correct Content-type header to send. I've tested and it resolves my particular problem.


stephenmcd:
That's great, thank you.

PR #1549, opened 2016-03-03
Google Chrome seems to interpret an empty response as <html><body></body></html>

zandeez:
Google Chrome seems to interpret an empty response as  <html><body></body></html> so forced content_type to text/plain to prevent Chrome handling it in this way.


stephenmcd:
Nice, thank you.

stephenmcd:
BTW, this has a test failure around style that we enforce. I've fixed this already, but next time please run the tests before submitting a PR.

Thanks again!

zandeez:
Not a problem. Worst bit: I use PyCharm so CTRL+SHIFT+L would have just applied pep8 anyway...

Issue #1542, opened 2016-02-23
Django 1.9.2 and Mezzanine 4.1.0


vladir:
Hi team,

  I'm having this issue after I started the upgrade of my project to `Django 1.9.2` and `Mezzanine 4.1.0`:

    (venv)vladir@um-girl:~/work/saxo-all/project-django1.9/saxo-publish$ python publish/manage.py runserver
    Unhandled exception in thread started by <function wrapper at 0x7fc5b08d2578>
    Traceback (most recent call last):
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/utils/autoreload.py", line 226, in wrapper
        fn(*args, **kwargs)
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/core/management/commands/runserver.py", line 109, in inner_run
        autoreload.raise_last_exception()
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/utils/autoreload.py", line 249, in raise_last_exception
        six.reraise(*_exception)
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/utils/autoreload.py", line 226, in wrapper
        fn(*args, **kwargs)
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/__init__.py", line 18, in setup
        apps.populate(settings.INSTALLED_APPS)
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/apps/registry.py", line 115, in populate
        app_config.ready()
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/django/contrib/admin/apps.py", line 22, in ready
        self.module.autodiscover()
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/mezzanine/boot/__init__.py", line 119, in autodiscover
        admin_site.lazy_registration()
      File "/home/vladir/work/saxo-all/project-django1.9/venv/local/lib/python2.7/site-packages/mezzanine/boot/lazy_admin.py", line 56, in lazy_registration
        getattr(AdminSite, name)(self, *args, **kwargs)
[...]

PR #1551, opened 2016-03-04
Fixing the issue django.contrib.admin.sites.NotRegistered: The model User is not registered.

vladir:
* To fix this https://github.com/stephenmcd/mezzanine/issues/1542


stephenmcd:
Thanks Vladir

PR #1553, opened 2016-03-04
Fix help text for Link page URL

gavinwahl:
The field is required, but the inherited help text says it's optional.


stephenmcd:
Thanks Gavin

Issue #1554, opened 2016-03-08
Why is the admin panel sidebar a .dropdown-menu?


GeoTob:
Just wondering why the admin panel sidebar is defined as a `.dropdown-menu` in `templates/admin/include/dropdown_menu.html`? It seems semantically wrong and breaks compatibility with other apps that try to use and style dropdown menus ?

I suspect that I may be missing an obvious reason for using dropdown-menu? If not, I'm happy to submit a patch later this week.

jerivas:
I think this is because several versions back, the admin menu was a 
dropdown menu (not a sidebar as you see today). I guess that only the 
styles got re-written without actually changing the class names.

Let's see what others have to say.

stephenmcd:
I don't have the answer off hand, but this should be posted to the mailing list instead, thanks.

stephenmcd:
I misread your issue as a question not a bug, and yes please submit a patch, thanks.

GeoTob:
i must admit it wasn't clearly worded @stephenmcd . Will submit something later this week.

PR #1557, opened 2016-03-10
Changed dropdown-menu to admin-sidebar for admin panel

GeoTob:
For for issue #1554 : The admin sidebar breaks once an app redefines styles for `.dropdown-menu`.

* changed class names from `.dropdown-menu` to `.admin-sidebar` and `.dropdown-menu-menu` to `.admin-app-menu`
* changed template name to `sidebar.html`
* changed template tag name to `admin_sidebar`
* modified other files accordingly

### DIFF
--- a/mezzanine/core/locale/ar/LC_MESSAGES/django.po
+++ b/mezzanine/core/locale/ar/LC_MESSAGES/django.po
@@ -562,7 +562,7 @@ msgstr "اقرأ المزيد"
 msgid "View site"
 msgstr "عرض الموقع"
 
-#: templates/admin/index.html:17 templates/admin/includes/dropdown_menu.html:4
+#: templates/admin/index.html:17 templates/admin/includes/sidebar.html:4
 msgid "Dashboard"
 msgstr "لوحة المعلومات"
 
--- a/mezzanine/core/locale/bg/LC_MESSAGES/django.po
+++ b/mezzanine/core/locale/bg/LC_MESSAGES/django.po
@@ -557,7 +557,7 @@ msgstr ""
 msgid "View site"
 msgstr ""
 
-#: templates/admin/index.html:17 templates/admin/includes/dropdown_menu.html:4
+#: templates/admin/index.html:17 templates/admin/includes/sidebar.html:4
 msgid "Dashboard"
 msgstr ""
 
--- a/mezzanine/core/locale/ca/LC_MESSAGES/django.po
+++ b/mezzanine/core/locale/ca/LC_MESSAGES/django.po
@@ -582,7 +582,7 @@ msgstr "Llegir més"
 msgid "View site"
 msgstr ""
 
-#: templates/admin/index.html:17 templates/admin/includes/dropdown_menu.html:4
+#: templates/admin/index.html:17 templates/admin/includes/sidebar.html:4
 msgid "Dashboard"
 msgstr "Taulell"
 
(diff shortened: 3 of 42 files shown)


stephenmcd:
I interpreted the original report (after re-opening it) as there being an actual bug, but now I see this is more of a semantic fix, right?

Given that, I feel like the amount of change here doesn't warrant the potential breakage it might cause others who somehow already depend on it (modifying styles, overriding templates, etc).

GeoTob:
`dropdown-menu` is Bootstrap's way of defining dropdown menus. When I bring in code that wants to restyle dropdown menus (for whatever reason, my trigger here was a required django-filer installation), the sidebar simply breaks.

We can shorten this PR to just replacing the `dropdown-menu` style, but I thought it best to fix the semantic issues as well. Feel free to take time testing this though, I have a local patch to work around.

stephenmcd:
Also if I recall correctly, `.dropdown-menu` originated in Mezzanine, so it isn't some pre-defined name that other things potentially use. 

If that's something you've experienced I'd put it down to coincidence unless I'm mistaken. The same bug could be made for any name used, unless it was `.mezzanine-super-unique-classname`, what's to stop another app from using '.admin-sidebar', and then what?

GeoTob:
@stephenmcd see: http://getbootstrap.com/components/#dropdowns

stephenmcd:
> dropdown-menu is Bootstrap's way of defining dropdown menus

But bootstrap isn't used in Mezzanine's admin, and this just highlights the point further - two entirely separate things depending on the same name by coincidence. Changing it to something else doesn't solve this in a general way.

stephenmcd:
To further add fuel, I'm pretty sure this existed in Mezzanine before bootstrap ever existed.

GeoTob:
If I install an app that tweaks form fields, I expect it to restyle forms, dropdowns etc. I also expect it to use semantically correct class names. If I install two such apps, I know that problems are likely due to name clashes. 

I do not expect it to break a sidebar navigation though, and in that case the css is one of the last places I'd look into for troubleshooting. A breaking sidebar because something uses a semantically sensible name for restyling form fields is a bug to me.

Don't think Mezzanine's age really counts as an argument here though  :)

stephenmcd:
You're arguing that when the class names in two different code bases collide, then one should change. 
Whether one or the other is more or less semantically correct isn't relevant - what happens when another third-party thing comes along that uses the new class name we change to? It's a silly argument to make for changing the class name in either piece of software, and the only silly data point relevant is whichever had the name first.

I'd suggest forking one of the conflicting projects, and changing the name in one of them, whichever is less pain - but expecting either project to change in order to resolve the conflict is silly.

Yes, `.dropdown-menu` isn't semantically as correct as it could be, but given the potential problems it could cause others reliant on it, I'm inclined to leave it as is. This is an entirely separate topic from names colliding with other packages though.

GeoTob:
I'm not _expecting_ your project to change, I'm suggesting a change. My argument is that we should minimise collisions where possible. Semantically correct names are one of the ways in which we can minimise collisions across applications (namespacing would be another).

If you actually want to style a dropdown menu, then go ahead and use `.dropdown-menu`. Every developer will expect this to collide with her own dropdown styles. But when you style headings, tables, body text, sidebars or whatever, using `.dropdown-menu` is a silly idea, invites collision with other applications, makes troubleshooting harder, and makes things harder to maintain by introducing potential points of failure.

I can totally see where you're coming from in terms of compatibility with other modules reliant on this, but I think this is an issue worth fixing. It's your project though, do as you like.

derek-adair:
Fixing CSS semantics like this should be done wholesale or not at all in my opinion.  Much easier to upgrade when you know to do it all.

a css prefix would go a long way.

PR #1559, opened 2016-03-10
Update blog share button links

solarissmoke:
Hi!

This small patch makes two changes to `blog_post_detail.html`:

For Twitter, use the new [web-intent API](https://dev.twitter.com/web/tweet-button/web-intent) which allows us to send the URL and text separately in structured format.

For Facebook, update the sharer URL and remove all parameters except the URL as [they are ignored](https://developers.facebook.com/x/bugs/357750474364812/) - it only let's you submit a URL which Facebook will crawl and fetch Open Graph content from.


stephenmcd:
Thanks Samir

Issue #1012, opened 2014-04-23
Task List: Tests that need to be written


jerivas:
# The reason for this list

As Mezzanine grows and expands to cover more and more use cases, every new commit has the potential of inadvertently breaking something somewhere in the code base.

This issue keeps track of the tests that need to be written to ensure certain features are not broken. The goal is to have a comprehensive test base to avoid nasty surprises on new releases.

If you contribute a new feature or fix a bug, please run the current tests, write some for your contributions and comment on this issue to have your tests added as completed. If you can't write the test, comment on this issue to have it added as pending. Lastly, you can contribute by tackling some of the pending tests of this list.
# The list
## Core
- [x] In the admin, the site drop-down must always be present when multiple sites exist. 7785481507df560cea5b517317ef2a620005ee5d
- [ ] In the admin, the language drop-down must always be present when i18n settings have been defined.
- [ ] In the admin inline items (both tabular and stacked) must be added via the "Add another" button
- [ ] In the admin the RichText and File fields added by dynamic inlines must work and save their values
## Pages
- [ ] In the admin, the page tree must allow re-ordering and nesting by drag-n-drop
## Blog
- [ ] `BlogPostForm` must set the hidden fields' initial values and widgets.
- [ ] `BlogPost` model's `get_absolute_url` method must build the correct url_name.
- [ ] `blog_post_list` view must filter posts based on tag, date, category and author.
- [ ] `BlogCategoryAdmin`'s `in_menu` should return `True` only if the `BlogCategory` model is in the `ADMIN_MENU_ORDER`.
- [ ] `PostsRSS` class should use the correct title and description based on the `Page`'s existence.
- [ ] `PostsRSS`'s must filter blog posts based on tag, category and author.
- [ ] `blog_tags` template tag should work correctly.
## Filebrowser (media library)
- [ ] Renaming and deleting files
- [ ] Uploading files (respecting the file size and type constraints)
- [ ] Searching for files
## Twitter
- [ ] Test all variations of Twitter queries. ASCII only and non-ASCII, User, Search, List, settings defined in Python vs the DB, Python 2 and Python 3. By my count that's 24 different paths that can occur - historically at least one of these is broken, and when fixing it, another will break.
## Generic
- [ ] Ensure comment values are correctly persisted through cookies and remain intact when pre-populating the comment form.

auvipy:
@stephenmcd having a mezzanine coverage report added on repo will be helpful

stephenmcd:
[...]

PR #1562, opened 2016-03-14
Add Coveralls coverage testing to CI, per #1012.

ryneeverett:
A collaborator just needs to go to coveralls.io, login via github oauth, and flip the switch to "On".


stephenmcd:
Awesome, thanks Ryne! I took a look in coveralls.io and it looks like it's already enabled there.

PR #1564, opened 2016-03-16
Value from Setting class must accept blank/null

tecbr:
conf\models.py

from
`value = models.CharField(max_length=2000)`

to:
`value = models.CharField(max_length=2000, null=True, blank=True)`

because this code from conf\forms.py causes exception, if object is created:
`setting_obj, created = Setting.objects.get_or_create(name=name)`

...and...why not accept blank values?


stephenmcd:
Works for me: 

```
>>> Setting.objects.get_or_create(name="test")
(<Setting: test: >, True)
```

PR #1566, opened 2016-03-18
Enable intersphinx

chmelevskij:
If referecing Django settings following syntax can be used:
:django:setting:`INSTALLED_APPS`

### DIFF
--- a/docs/content-architecture.rst
+++ b/docs/content-architecture.rst
@@ -269,7 +269,7 @@ custom :class:`.Page` models and are then called inside the
 :class:`.Page` instance. A Page Processor will always be passed two arguments
 - the request and the :class:`.Page` instance, and can either return a
 dictionary that will be added to the template context, or it can return
-any of Django's `HttpResponse`` classes which will override the
+any of Django's ``HttpResponse`` classes which will override the
 :func:`mezzanine.pages.views.page` view entirely.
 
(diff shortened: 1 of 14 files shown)


docs/content-architecture.rst
```diff
@@ -269,12 +269,12 @@ custom :class:`.Page` models and are then called inside the
 :class:`.Page` instance. A Page Processor will always be passed two arguments
 - the request and the :class:`.Page` instance, and can either return a
 dictionary that will be added to the template context, or it can return
-any of Django's ``HttpResponse`` classes which will override the
+any of Django's `HttpResponse`` classes which will override the
```
stephenmcd: This looks wrong

stephenmcd:
Cool idea! Could you apply it to the rest of the documentation?

docs/content-architecture.rst
chmelevskij: It is wrong... Probably just something left over from testing.

chmelevskij:
Will do that as well. Should I just close this pull request and make a new one afterwards? I just thought it could be useful for someone writing the docs as well.

stephenmcd:
You could just add to this current PR if you like. Thanks a lot!

pedromorgan:
I was about to make an issue of interphinx, but its already here

chmelevskij:
All that needs to be done is to add the actual tags.....

PR #1569, opened 2016-03-21
Clear TEST_COLLATION deprecation warning

dsanders11:
This patch gets rid of the deprecation warning when using MySQL:

`RemovedInDjango19Warning: In Django 1.9 the TEST_COLLATION connection setting will be moved to a COLLATION entry in the TEST setting`

The database test settings moved under a `TEST` parent setting with [Django 1.7](https://docs.djangoproject.com/en/1.7/ref/settings/#test) so it is supported by all versions of Django that Mezzanine supports. Should make the line of code compatible with 1.9.

### DIFF
--- a/mezzanine/utils/conf.py
+++ b/mezzanine/utils/conf.py
@@ -214,7 +214,7 @@ def set_dynamic_settings(s):
                 s["DATABASES"][key]["NAME"] = db_path
         elif shortname == "mysql":
             # Required MySQL collation for tests.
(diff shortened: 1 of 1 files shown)


dsanders11:
Build failure was a PEP8 fail, I keep forgetting this project enforces lines at 80 chars.

dsanders11:
Anything need to be done here for acceptance? Shouldn't have any functional impact other than clearing the deprecation warning.

stephenmcd:
Sorry I'd forgotten this, thanks.

PR #1572, opened 2016-03-23
Feature/configurable profile inline

mmansour:


mezzanine/accounts/__init__.py
```diff
@@ -75,6 +75,20 @@ def get_profile_form():
                                    settings.ACCOUNTS_PROFILE_FORM_CLASS)
 
 
+def get_profile_inline_form():
```
jerivas: I find the name of this function misleading. It currently implies that a form class is returned, not an admin class. I suggest we rename it to `get_profile_inline_admin()`.

mezzanine/accounts/admin.py
```diff
@@ -45,20 +46,15 @@ def save_model(self, request, obj, form, change):
...
 
-        def get_min_num(self, request, obj=None, **kwargs):
-            """This causes profile forms to be shown when editing but hidden
-            when creating. If min_num is fixed at 1, Django's initial user
-            creation form fails if the profile model has a required field."""
-            return 0 if obj is None else 1
 
-    UserProfileAdmin.inlines += (ProfileInline,)
+try:
+    UserProfileAdmin.inlines += (get_profile_inline_form(),)
```
jerivas: Wondering if we even need to create a wrapper function for this. `import_dotted_path()` already handles exceptions with sufficient output to debug if something goes wrong. Maybe we can call it here directly and drop this function altogether?

mezzanine/accounts/defaults.py
```diff
@@ -56,6 +56,14 @@
 )
 
 register_setting(
+    name="ACCOUNTS_PROFILE_INLINE_CLASS",
+    description=_("Dotted package path of the profile inline "
```
jerivas: IMHO this could explain a bit better that this appears in the User admin. Something like "Dotted package path and class name of the profile inline admin that will be added to the User admin, when `mezzanine.accounts` is installed and `settings.AUTH_PROFILE_MODULE` is defined."

mezzanine/accounts/__init__.py
mmansour: I agree. Done

mezzanine/accounts/admin.py
mmansour: Seems reasonable to me. I don't see a  reason we need trap and log for `ProfileNotConfigured` anymore. But I could easily be overlooking something. I will defer to you all on that one.

mezzanine/accounts/defaults.py
mmansour: Done

mezzanine/accounts/admin.py
jerivas: Cool, let's see what @stephenmcd has to say.

stephenmcd: Agree with @jerivas, function seems unnecessary, especially since the default should never fail since the setting will default to something that always works

stephenmcd: Also it only gets called in one spot, so there's no notion of reusability

mmansour: Cool, thanks. Removed.

stephenmcd: Looks like it's still there.

PR #1574, opened 2016-03-31
Only autofocus visible fields with Html5Mixin

dsanders11:
Check if a widget is visible before adding the `autofocus` attribute. In Cartridge where form fields are hidden between checkout steps the Html5Mixin was adding `autofocus` to hidden fields, which is both not very useful, and fails HTML validation.


stephenmcd:
Nice one, thanks David.

PR #1577, opened 2016-04-02
Replace all _meta.fields with _meta.get_fields.

go8ose:
According to https://docs.djangoproject.com/en/1.8/ref/models/meta/ to
access all the fields of the _meta object you should call get_fields(). This
is new with Django 1.8. The _meta.fields is now a property with a docstring
saying it's Private API intended only to be used by Django itself.


stephenmcd:
Looks good. Have you verified that they return the same results?

PR #1579, opened 2016-04-03
Clarified thumbnail image name

wimfeijen:


stephenmcd:
Breaks the build, so prob not worth it, thanks anyway.

PR #1584, opened 2016-04-16
Ensure csrftoken cookie is set inside nevercache tag

dsanders11:
I'm currently hunting down occasional CSRF failures for users and noticed that Mezzanine isn't setting the CSRF cookie when it should for logged in users. It's not the cause for the issues I'm tracking, but it is a potential issue for other setups.

When the nevercache tag is used, any `{% csrf_token %}` inside the tag will not trigger a set cookie for the `csrftoken` cookie because the `CsrfViewMiddleware` has already run by the time `UpdateCacheMiddleware` renders the contents of the nevercache tag. Since the standard Mezzanine template include `core/templates/includes/form_fields.html` wraps the `{% csrf_token %}` with a nevercache tag this affects most forms being rendered by Mezzanine. It can be observed by logging into the demo site and going to 'Update Profile'. If you inspect the HTTP response there is no Set-Cookie for `csrftoken`.

Generally this isn't a problem because the cookie is set on login and has an expiration time of 1 year. Clearing cookies will also clear the session cookie and log the user out, so it's hard to end up in a situation where the session cookie is set (so the user is logged in) but not the CSRF cookie. In theory if the CSRF token is not refreshed within a year it will start failing, but it's unlikely for a user to be logged in for a year. This length is however [configurable](https://docs.djangoproject.com/en/1.9/ref/settings/#csrf-cookie-age) and there may be use cases (the docs mention IE problems) where setting it to `None` for a session-length cookie is needed, which could result in the situation where the user is logged in but the CSRF token cookie is not set after the the user closes their browser and reopens it.

Another possible case this would be problematic would be if a server is setup to use HTTP Basic Auth and Django is using a [Remote User Auth](https://docs.djangoproject.com/es/1.9/howto/auth-remote-user/). In that case there is no login form to cause the CSRF token to be set so the first time the user tries to use a form with this bug it would fail.

The pull request simply runs the `CsrfViewMiddleware` again for the response to set the cookie if `{% csrf_token %}` was used in a nevercache tag. There are no observable side effects from running it twice in the case that the cookie has already been set by a `{% csrf_token %}` not in a nevercache tag.

### DIFF
--- a/mezzanine/core/middleware.py
+++ b/mezzanine/core/middleware.py
@@ -173,6 +173,15 @@ class UpdateCacheMiddleware(object):
         if hasattr(request, '_messages'):
             # Required to clear out user messages.
             request._messages.update(response)
+        if csrf_token:
+            # Response needs to be run-through the CSRF middleware again so
+            # that if there was a {% csrf_token %} inside of the nevercache
+            # the cookie will be correctly set for the the response
+            csrf_mw_name = "django.middleware.csrf.CsrfViewMiddleware"
+            if csrf_mw_name in settings.MIDDLEWARE_CLASSES:
+                request.csrf_processing_done = False
+                csrf_mw = CsrfViewMiddleware()
+                csrf_mw.process_response(request, response)
         return response
 
 
--- a/mezzanine/core/middleware.py
+++ b/mezzanine/core/middleware.py
@@ -173,15 +173,14 @@ class UpdateCacheMiddleware(object):
         if hasattr(request, '_messages'):
             # Required to clear out user messages.
             request._messages.update(response)
-        if csrf_token:
-            # Response needs to be run-through the CSRF middleware again so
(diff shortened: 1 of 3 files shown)


dsanders11:
Confirmed that the easiest way to reproduce the issue this fixes (I've modified the pull request to remove an over limiting check for an existing CSRF token) is to put `CSRF_COOKIE_AGE = None` in your settings, log in, close the browser (I used FireFox, Chrome wasn't keeping me logged in between closing and opening) and then immediately go to the 'Update Profile' page and try to submit that form. It will fail with a CSRF token cookie not set error. Applying this patch makes it work as expected.

dsanders11:
Added a testcase, although I don't love the hack I had to put around `nevercache`. Since it gets set at module load time I had to find a way to force it to be the caching version for that one specific test and then go back to being the no-op version.

stephenmcd:
Great work David, thank you.

PR #1592, opened 2016-05-02
Check for backend-dependent template as well in template middleware

dsanders11:
Since Django 1.8 it's been possible to use a backend-dependent template with `SimpleTemplateResponse` and `TemplateResponse`, and in Django 1.10 the ability to not use a backend-dependent template is going away (deprecated since 1.8). For now check for either `Template` type, and eventually drop the check in favor of only checking backend-dependent template once only 1.10+ is supported.


AlexHill:
Hi David,

Nice catch.

I've changed the patch a bit to accommodate backends other than the default – have a look at commit e60d1e998c90565cab834301d86e7dd574a7f1c8 and let me know what you think.

Alex

stephenmcd:
Does 4975881c07df37d252cf2f36842862d75c9012f7 negate this PR?

PR #1596, opened 2016-05-08
Admin app list improvements

ryneeverett:
See commit messages for detail.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -549,6 +549,15 @@ def admin_app_list(request):
     for (model, model_admin) in admin.site._registry.items():
         opts = model._meta
         in_menu = not hasattr(model_admin, "in_menu") or model_admin.in_menu()
+        if hasattr(model_admin, "in_menu"):
+            import warnings
+            warnings.warn(
+                'ModelAdmin.in_menu() has been replaced with '
+                'ModelAdmin.has_module_permission(request). See '
+                'https://docs.djangoproject.com/en/stable/ref/contrib/admin/'
+                '#django.contrib.admin.ModelAdmin.has_module_permission.',
+                DeprecationWarning)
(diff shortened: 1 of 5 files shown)


ryneeverett:
Heh, might want to change the "COVERAGE DECREASE THRESHOLD FOR FAILURE" setting in coveralls.io from the default of "0".

AlexHill:
@ryneeverett thanks, looks good.

mezzanine/core/templatetags/mezzanine_tags.py
```diff
@@ -559,14 +559,13 @@ def admin_app_list(request):
                 DeprecationWarning)
         in_menu = in_menu and model_admin.has_module_permission(request)
         if in_menu and request.user.has_module_perms(opts.app_label):
-            perms = model_admin.get_model_perms(request)
             admin_url_name = ""
-            if perms["change"]:
+            if model_admin.has_change_permission(request, model):
```
gradel: As far as I understand it, has_change_permission takes an object as second argument, not a model. If you want to check permission for all objects of type model, leave out the second argument.

ryneeverett: @gradel I think you're right. I did test this and it seems like this should have thrown an error.

gradel: @ryneeverett I have django-filer integrated and this is the only app, which really throws an error here.

ryneeverett: Oddly enough coveralls says this line runs 6x in tests. https://coveralls.io/builds/6921715/source?filename=mezzanine%2Fcore%2Ftemplatetags%2Fmezzanine_tags.py#L591

ryneeverett: I'm guessing that if the user has permissions for all objects of type model, the object is simply ignored. That would explain why I can pass any arbitrary object and it returns `True`.

PR #1610, opened 2016-06-07
Remove leading slash from BLOG_SLUG url pattern

zruu:
Gives error "?: (urls.W002) Your URL pattern '^/' has a regex beginning with a '/'. Remove this slash as it is unnecessary."

Mezzanine 4.1.0
Django 1.9.6
Python 3.4.4
MySQL 5.7.10
Darwin 15.5.0

### DIFF
--- a/mezzanine/urls.py
+++ b/mezzanine/urls.py
@@ -67,7 +67,7 @@ if "mezzanine.accounts" in settings.INSTALLED_APPS:
 # Mezzanine's Blog app.
 blog_installed = "mezzanine.blog" in settings.INSTALLED_APPS
 if blog_installed:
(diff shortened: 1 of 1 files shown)


stephenmcd:
Was already fixed in fdfafa47e038a397653f5afe898b84fa7402aa2e

zruu:
Alright, I tested it again with BLOG_SLUG in settings to an empty string and it gives me a %2F after which is a leading slash. If I set BLOG_SLUG to 'blog' or anything else it works perfectly.

stephenmcd:
Thanks for the info, I see the problem now. The previous fix wasn't quite right though, I've pushed the correct fix in 5a8e38cdc44d693320729a2e4baf76051ed2541b

PR #1612, opened 2016-06-09
Content Typed Mixins

ryneeverett:
See stephenmcd/cartridge#293 for motivation.


ryneeverett:
A couple things perhaps worth considering:

- The `mezzanine/pages/templates/admin/shop/product/change_list.html` and `cartridge/shop/templates/admin/shop/product/change_list.html` templates still have some duplication. There isn't an obvious place to put a `custom_content_types/change_list.html` template since utils isn't an app and perhaps it's not worth worrying about.
- From stephenmcd/cartridge#57:

> the names of `content_model` and `get_content_model()` are taken directly from Mezzanine and don't really make sense here. I think they should either be changed, or this functionality should be factored out into a mixin in Mezzanine core.

It isn't clear to me whether @AlexHill was thinking the names should be changed *if* the functionality is "factored out into a mixin in Mezzanine core." I'm not that concerned about the names but if this is something we want to address let's do it now.

stephenmcd:
> It isn't clear to me whether @AlexHill was thinking the names should be changed if the functionality is "factored out into a mixin in Mezzanine core."

Here's my take: when the code was being duplicated, something like `get_product_model` made much more sense. But now we need to consider a common name. `content` in Mezzanine has always been a fairly ambiguous thing, a notion of "content types" gets used on the mailing list and throughout the docs to mean custom page types, and `get_content_model` really should have been `get_page_model` from the start, but that bridge is crossed.

So now we're looking at a common name to represent the idea that there's an underlying subclass instance backing the model, or a "content type". In fact I think abstracting this out as you've done for Cartridge, the notion of a "content type" gets a little less ambiguous, since it's no longer a confusing way of saying "page type" - it actually represents something more abstract. 

I think "content" is a good name for this, and it means not changing anything! Open to better suggestions though.

mezzanine/pages/admin.py
```diff
@@ -105,24 +66,16 @@ def add_view(self, request, **kwargs):
         return super(PageAdmin, self).add_view(request, **kwargs)
 
     def change_view(self, request, object_id, **kwargs):
-        """
-        For the ``Page`` model, check ``page.get_content_model()``
-        for a subclass and redirect to its admin change view.
-        Also enforce custom change permissions for the page instance.
-        """
+        """ Enforce custom permissions for the page instance. """
```
stephenmcd: Nit: Can we keep docstring format consistent with the rest of the project? We always put the triple quotes on separate lines, even for one-line docstrings. Whether it's right is arguable, but it's consistent.

mezzanine/utils/custom_content_types.py
```diff
@@ -0,0 +1,152 @@
...
+from django.core.urlresolvers import NoReverseMatch
+from django.db import models
+from django.http import HttpResponseRedirect
+from django.shortcuts import get_object_or_404
+
+from mezzanine.utils.models import base_concrete_model
+from mezzanine.utils.urls import admin_url
+
+
+class CustomContentTypes(models.Model):
```
stephenmcd: Not a fan of "Custom" in the name, I'd prefer something like "ContentTyped". Thoughts?

Also I don't think this belongs under utils, better off in mezzanine.core.{models,admin} - what's in there already is generally abstract/util-ish and used throughout Mezzanine.

mezzanine/utils/custom_content_types.py
```diff
@@ -0,0 +1,152 @@
...
+        """ Return all subclasses of the concrete model.  """
+        concrete_model = base_concrete_model(CustomContentTypes, cls)
+        return [m for m in apps.get_models()
+                if m is not concrete_model and issubclass(m, concrete_model)]
+
+    def get_content_model(self):
+        return getattr(self, self.content_model, None)
+
+
+class CustomContentTypesAdmin(object):
```
stephenmcd: Nit: empty line here, again possibly wrong but for consistency.

mezzanine/utils/custom_content_types.py
```diff
@@ -0,0 +1,152 @@
...
+                pass
+
+            fields = self.model._meta.fields + self.model._meta.many_to_many
+            for field in reversed(fields):
+                if field.name not in exclude_fields and field.editable:
+                    if not hasattr(field, "translated_field"):
+                        self.fieldsets[0][1]["fields"].insert(3, field.name)
+
+    @property
+    def base_conrete_modeladmin(self):
```
stephenmcd: conrete -> concrete (also elsewhere in the code)

mezzanine/utils/custom_content_types.py
ryneeverett: > Not a fan of "Custom" in the name, I'd prefer something like "ContentTyped". Thoughts?

I didn't like the name either but couldn't think of anything better. "ContentTyped" it is.

ryneeverett:
I should probably pull the js necessary for `content_typed/change_list.html` into a separate file and add it to the include so the issue is taken care of on the Cartridge side.

ryneeverett:
Ok, so [the `toggleID` function in page_tree.js](https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/pages/static/mezzanine/js/admin/page_tree.js#L12) is used by this and the actual tree ordering code. We could duplicate that or does mezzanine have a js utils module I could throw it in?

ryneeverett:
The duplication isn't that bad since `opened` is always `true` so the function can be pruned down quite a bit.

mezzanine/utils/custom_content_types.py
stephenmcd: Can we put them into `mezzanine.core.models` and `mezzanine.core.admin`? Don't think a special module is warranted.

mezzanine/core/static/mezzanine/js/admin/content_typed.js
```diff
@@ -0,0 +1,28 @@
...
+    var toggleID = function(id) {
+        // Add or remove the page ID from the cookie IDs string.
+        var index = $.inArray(id, ids.split(','));
+        if (index == -1) {
+            if (ids) {ids += ',';}
+            ids += id;
+        }
+        document.cookie = cookie + '=' + ids + '; path=/';
+    };
+
```
stephenmcd: This still doesn't seem right. `toggleID` has nothing to do with the new "content typed" stuff, it's purely for pages and page_tree.js

The JS for content typed stuff should literally be a one-line function that probably doesn't even warrant its own file, eg:

`location.href = this[this.selectedIndex].value;`

The tricky part is how page_tree.js then integrates with that so it can run `toggleID` - I'm not sure about the best way to do that, and what order things occur in with jQuery when dealing with multiple event handlers, but there must be some sane way to deal with that.

mezzanine/pages/static/mezzanine/js/admin/page_tree.js
```diff
@@ -59,21 +59,13 @@ jQuery(function($) {
...
         // the new branch will be visible directly after saving.
         var id = $(this).attr('id');
         if (id) {
             toggleID(true, id.split('-')[1]);
         }
-        var addUrl = this[this.selectedIndex].value;
-        if (addUrl) {
-            location.href = addUrl;
-        }
-        this.selectedIndex = 0;
```
ryneeverett: The purpose of the conditional `if (addUrl)` and `this.selectedIndex = 0` is unclear to me. The only `<option>` that doesn't have a `value` is the first one (`<option value="">{% trans "Add" %} ...</option>`) which is always selected on page load so it cannot trigger the `change` handler.

mezzanine/pages/static/mezzanine/js/admin/page_tree.js
```diff
@@ -59,21 +59,13 @@ jQuery(function($) {
...
-        return true;
```
ryneeverett: This line doesn't seem to do anything. http://stackoverflow.com/a/4379435/1938621

mezzanine/core/static/mezzanine/js/admin/content_typed.js
ryneeverett: > The tricky part is how page_tree.js then integrates with that so it can run toggleID - I'm not sure about the best way to do that, and what order things occur in with jQuery when dealing with multiple event handlers, but there must be some sane way to deal with that.

Oddly enough, it seems that all handlers are called regardless of the redirect so the order doesn't matter. That is, if the `location.href` is changed in the first handler called, it doesn't take effect until after the second handler is called.

stephenmcd: Cool. After first scanning the code I thought the ordering of the two events was important, but looking at it now I see it isn't. Not so tricky then. :-)

mezzanine/pages/static/mezzanine/js/admin/page_tree.js
stephenmcd: It's only a guess but it's a confident one - I think that was to handle the situation when the user had been directed to the "add" interface, hit the back button, and the browser had the tree page in cache/memory, and keeps the state of the select list. So the select list would have a page type selected, and there'd be no way to trigger the JS to run again other than manually selecting nothing (eg selectedIndex = 0) and selecting it again.

stephenmcd: BTW if this doesn't actually do what I think it does, we can remove it.

Same for the return statement.

stephenmcd: Sorry, I totally missed the fact you've removed the code already. :-) So assuming it doesn't break what I described, let's go with what you've done.

stephenmcd: Actually I've just tested this now and the back button problem still exists, but I'll fix that myself when merging this.

Thanks for all your work on this and on addressing all the feedback! 

I'll look at the Cartridge side next.

PR #1615, opened 2016-06-11
unicode method for form.fields

Roastmaster:
As django templates and models primarily translate all strings to unicode.  This may be preferred and less confusing over the status quo u'field_01', u'field_02', etc.


jerivas:
The `__str__` method is used by Django when the `@python2_unicode_compatible` decorator is used. No need to define `__unicode__`. https://docs.djangoproject.com/en/1.9/ref/utils/#django.utils.encoding.python_2_unicode_compatible

PR #1618, opened 2016-06-16
dynamically set current rating if it exists

mgaitan:
this patch initialize the rating form with the proper value if authenticated user already rate a content.


stephenmcd:
Awesome, thank you.

PR #1626, opened 2016-07-08
Added a link

rosehosting:
We added our tutorial on how to install Mezzanine on Debian https://www.rosehosting.com/blog/install-mezzanine-cms-on-a-debian-vps/


stephenmcd:
Don't think this belongs in the docs, thanks anyway.

PR #1627, opened 2016-07-11
Add form media to Form pages.

ryneeverett:
My use case is adding form assets to a widget used by a field included
in the FORMS_EXTRA_FIELDS setting. I don't think one should have to
override this template to do this and overextension doesn't seem to work
on content-typed templates.


stephenmcd:
Nice, thanks Ryne.

PR #1631, opened 2016-07-20
Fix has_change_permission invocation.

ryneeverett:
As @gradel pointed out, the optional "object" argument takes a model
instance, not a model.

I'm guessing the reason this hasn't been blowing up with errors is that
the object argument is ignored if the user has change permissions to all
objects in the model.


stephenmcd:
Thanks Ryne.

PR #1635, opened 2016-07-24
Django 1.10 compatibility fixes

### LINKED ISSUE #1619: Django 1.10 support
- [x] `django.shortcuts.render_to_response` no longer supports `context_instance` arg (affects filebrowser)
- [x] `LOGOUT_URL` setting no longer defined
- [x] `SearchableManager.contribute_to_class` no longer achieves workaround for abstract models being disallowed access to managers
- [x] `OptParser` is no longer used by Django for management commands (refs to `Command.option_list` fail)
- [x] `NoArgsCommand` is removed from Django
- [x] Django's `i18n_patterns()` can return a tuple in 1.10 (see django/django#6968)
- [x] redirect loop in mezzanine.pages.tests.PagesTests.test_login_required

AlexHill:
Miscellaneous Django 1.10 fixes for review.

### DIFF
--- a/mezzanine/bin/runtests.py
+++ b/mezzanine/bin/runtests.py
@@ -6,6 +6,7 @@ import shutil
 import sys
 
 import django
+from django.core.management import call_command
 
 
 def main(package="mezzanine"):
@@ -63,7 +64,9 @@ PASSWORD_HASHERS = ('django.contrib.auth.hashers.MD5PasswordHasher',)
     django.setup()
 
     from django.core.management.commands import test
-    sys.exit(test.Command().execute(verbosity=1))
+    if django.VERSION < (1, 10):
+        sys.exit(test.Command().execute(verbosity=1))
+    sys.exit(call_command(test.Command(), verbosity=1))
 
 
 if __name__ == "__main__":
--- a/mezzanine/core/managers.py
+++ b/mezzanine/core/managers.py
@@ -1,4 +1,6 @@
 from __future__ import unicode_literals
+
+import django
 from future.builtins import int, zip
 
 from functools import reduce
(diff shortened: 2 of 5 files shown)


mezzanine/bin/runtests.py
```diff
@@ -63,7 +64,9 @@ def cleanup_test_settings():
     django.setup()
 
     from django.core.management.commands import test
-    sys.exit(test.Command().execute(verbosity=1))
+    if django.VERSION < (1, 10):
+        sys.exit(test.Command().execute(verbosity=1))
+    sys.exit(call_command(test.Command(), verbosity=1))
```
AlexHill: Defaults weren't being set properly in 1.10 when calling the command directly so I was getting an error running the tests. Might be connected to dropping optparse support, but I didn't dig too deep really.

mezzanine/core/managers.py
```diff
@@ -19,6 +21,34 @@
...
+                    "'%s.%s' has been swapped for '%s'" % (
+                        cls._meta.app_label,
+                        cls._meta.object_name,
+                        cls._meta.swapped,
+                    )
+                )
+
+            return cls._meta.managers_map[self.manager.name]
+
+
```
AlexHill: This seems to work but it's very ugly. It might be worth creating a dedicated search API which doesn't use the Manager convention, and to deprecate this?

mezzanine/utils/conf.py
```diff
@@ -19,9 +20,17 @@ class SitesAllowedHosts(object):
...
+                self._hosts = [
+                    s.domain.split(":")[0] for s in Site.objects.all()
+                ]
+            except OperationalError:
+                return iter([])
         return iter(self._hosts)
 
+    def __add__(self, other):
+        return list(self) + other
+
```
AlexHill: Django 1.10 appends ['testserver'] to `ALLOWED_HOSTS` while running tests, and does so before migrations have been run. So SitesAllowedHosts needs to handle addition with a list, and its database table being missing.

AlexHill:
There's a bug in the 1.10 release candidate which thoroughly breaks Mezzanine - see django/django#6933. Until that's merged the tests won't run, but if you clone that branch and bump the version in `setup.py` to `< 1.12`, the test suite passes with the exception of `mezzanine.pages.tests.PagesTests.test_login_required`.

mezzanine/core/managers.py
stephenmcd: I feel like it'd be better to hold off on deprecating this until there's literally no way to keep it working.

Perhaps this is the last hack we'll need and it just works from here on in.

stephenmcd:
Awesome Alex, I've merged it. I also updated the list in #1619

stephenmcd:
BTW I couldn't get the tests to fail other than the login redirect loop problem, which I fixed just now.

AlexHill:
Huh. Without django/django#6933, any test that involves a response being returned from PageMiddleware is broken for me, so I held off adding 1.10 to the Travis configuration.

stephenmcd:
Maybe I have something defined in my local_settings module that's changing the behaviour.

PR #1636, opened 2016-07-25
Replace usage of optparse with argparse

AlexHill:
Very straightforward replacement since our use of `optparse` wasn't complicated. I removed the `default` argument for options with `action="store_true"` or `action="store_false"`.

### DIFF
--- a/mezzanine/core/management/commands/createdb.py
+++ b/mezzanine/core/management/commands/createdb.py
@@ -1,15 +1,12 @@
 from __future__ import print_function, unicode_literals
 from future.builtins import int, input
 
-from optparse import make_option
 from socket import gethostname
 
-from django import VERSION
 from django.core.management.base import BaseCommand, CommandError
 from django.contrib.auth import get_user_model
 from django.contrib.sites.models import Site
 from django.core.management import call_command
-from django.core.management.commands import migrate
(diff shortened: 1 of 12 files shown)


mezzanine/core/management/commands/createdb.py
```diff
@@ -26,25 +23,16 @@ class Command(BaseCommand):
...
-    def __init__(self, *args, **kwargs):
-        """
-        Adds extra command options (executed only by Django <= 1.7).
-        """
-        super(Command, self).__init__(*args, **kwargs)
-        if VERSION[0] == 1 and VERSION[1] <= 7:
-            self.option_list = migrate.Command.option_list + (
-                make_option("--nodata", action="store_true", dest="nodata",
-                    default=False, help="Do not add demo data."),)
-
```
AlexHill: This copies the arguments from the `migrate` command, which the (existing) add_arguments() method doesn't. This has been fine for the last few versions so I'm assuming it's unnecessary.

stephenmcd: Yeah I think the min we support right now is 1.8, so this is dead code.

stephenmcd:
Thanks Alex!

Issue #1637, opened 2016-07-26
Exceptions in get_ascendants


flo-dhalluin:
for some reason, I am not sure how to reproduce though (happens only with the prod database..) 
I get the following exception when get ascendents of a page in a template : 

... 

`  File "/opt/mymedecine/venv/lib/python3.4/site-packages/django/template/base.py", line 850, in resolve
    value = self._resolve_lookup(context)
  File "/opt/mymedecine/venv/lib/python3.4/site-packages/django/template/base.py", line 913, in _resolve_lookup
    current = current()
  File "/opt/mymedecine/venv/lib/python3.4/site-packages/mezzanine/pages/models.py", line 121, in get_ascendants
    self._ascendants = pages[0]._ascendants
`

ryneeverett:
The traceback is missing the actual exception.

AlexHill:
Does this happen with any page or just some? Only pages with parents? Only pages with custom slugs?

This must be happening due to some particular arrangement of your page tree and slugs, so we'll really need more detail to work out what's going on here.

Your fix might be all that's needed, but until we know what's causing the problem we can't know that for sure.

flo-dhalluin:
Ok, I finally got some time to investigate this one. pheww 

I was calling the method, on a page that's **not on the current site** . 
I have something like : 

```
class ItemViews(models.Model):
    # ... 
    item = models.ForeignKey(Page) # this is not filtered by current site.
```

And in my code ( and the production instance is of course multi site), I somehow retrieve all ItemViews for current user, and try to call get_ascendants on them, which triggers the exception, if the related page is not on the current site. I fixed the initial issue by having a proper custom manager for my ItemsView model that filters only items whose attached Page is on the current site, but I believe my proposed PR is still valid, we are calling [0] on a list we know may be empty ...

AlexHill:
@flo-dhalluin this is now fixed in master. Thanks for reporting the bug and for taking the time to investigate.

PR #1638, opened 2016-07-26
check if pages list is not empty in get_ascendants

flo-dhalluin:
fix for #1637

### DIFF
--- a/mezzanine/pages/models.py
+++ b/mezzanine/pages/models.py
@@ -113,13 +113,13 @@ class Page(BasePage, ContentTyped):
             # run it to see if we can retrieve all parents in a single
             # query, which will occur if the slugs for each of the pages
             # have not been customised.
+            self._ascendants = []
             if self.slug:
                 kwargs = {"for_user": for_user}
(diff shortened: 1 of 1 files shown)


stephenmcd:
This fix is incorrect given that you admittedly don't know what the cause is, and it's probably masking some other error.

Why don't you take a look at the data that's causing the exception to occur, and reporting back on what you find.

mezzanine/pages/models.py
```diff
@@ -113,13 +113,13 @@ def get_ascendants(self, for_user=None):
             # run it to see if we can retrieve all parents in a single
             # query, which will occur if the slugs for each of the pages
             # have not been customised.
+            self._ascendants = []
             if self.slug:
                 kwargs = {"for_user": for_user}
                 pages = Page.objects.with_ascendants_for_slug(self.slug,
```
flo-dhalluin: with_ascendants_for_slug() may return an empty list 

mezzanine/pages/managers.py 
https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/pages/managers.py#L71

PR #1640, opened 2016-08-01
check if pages list is not empty in get_ascendants

flo-dhalluin:
See comments on the reported issue. 

This happens when we call get_ascendants() on a page that's not on the current site in my case.


AlexHill:
Closing in favour of #1654 which includes a test.

PR #1646, opened 2016-08-12
Fix 'N comments' pt_BR translation

nossila:
Singular and plural pt_BR translation missing for '1 comment' and 'N comments'


stephenmcd:
Thanks Alisson

Issue #1565, opened 2016-03-17
Keyword field loses current value when Page form is invalid


stephenmcd:
As per: https://groups.google.com/forum/#!topic/mezzanine-users/7GD4ETP_Oyo

I've reproduced this, not sure what the cause is.

XuryaX:
Pull Requested for Fix.

PR #1648, opened 2016-08-13
Keyword Field Update for Validation Errors - Issue #1565

XuryaX:
[Issue #1565]: Keyword field loses current value when Page form is invalid

Fix: 
Added conditions for assigning keywords to the keyword widget, when a validation error occurs.


XuryaX:
Can anyone tell me why the build is failing? There are only some minor changes and I have also checked for PEP8 Style guidelines. the pep8 module does not tell me any styling guidelines missed.

AlexHill:
Hi Shauryadeep,

To check my understanding: when the form doesn't validate, the KeywordsWidget is passed the submitted *form*'s value for this field, rather than the *model*'s value. The current `decompress()` method only expects the model's value, which is a QuerySet, and so does the wrong thing when presented with a string from the form. Is that right?

It seems strange to me that we'd get these different datatypes being passed to decompress, but maybe that's just a side-effect of `KeywordField` being pretty advanced.

It would be great if you could include a test with this PR.

Thanks for contributing!

XuryaX:
Yes Alex, what you understood is exactly what is happening.
To my understanding, there are three cases are because

1. When a new BlogPost/RichTextPage is created -> There is no particular instance attached to the form, and this when rendered the data passed to `decompress `is `None`.

2. When an existing BlogPost/RichTextPage is opened -> The form is instantiated with the model instance,
and when the form is rendered, it assigns the value from instance to the field.  Here a query set of `Assigned Keyword` is passed to decompress.

3. When a validation error occurs, the form instantiates itself from the POST data. This is probably because the form should not query the database for the model instance.  Since validation error could be faced even when creating new RichTextPage. And also, the invalidated form should have the modified values of the form , even if an existing model record exists.
The value is extracted from the `get_value_from_datadict `method of the widget which extracts the keyword id string from `keywords_0` field of the POST data. And at this time, the decompress method gets a comma separated string of keyword pks.


I observed an inconsistent behavior while working on this. As per the approach followed for keywords in mezzanine that there should be no unused keywords. But if a validation fails on a new post and someone discards the post, the keywords would still be present in the DB with no assigned keyword attached to it. Although this behavior will probably be not faced by anyone unless purposefully done.

XuryaX:
@AlexHill Apparently this code was never pulled. I do understand that the code might not meet the code quality that you are looking for. If that is the case please do let me know, so that if I contribute in the future, that it would be accepted.

stephenmcd:
Sorry for the delay - merging this now.

Thanks a lot for the fix.

PR #1649, opened 2016-08-14
Stop using deprecated template settings.

ryneeverett:
- TEMPLATE_DIRS
- TEMPLATE_LOADERS

I think these must have been actually removed in django-1.10.0 because
they're throwing AttributeError's now.

### DIFF
--- a/mezzanine/template/loader_tags.py
+++ b/mezzanine/template/loader_tags.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 from future.builtins import map
 
 import os
+from itertools import chain
 
 from django.template import Template, TemplateSyntaxError, TemplateDoesNotExist
 from django.template.loader_tags import ExtendsNode
@@ -58,8 +59,11 @@ class OverExtendsNode(ExtendsNode):
         if context_name not in context:
             context[context_name] = {}
         if name not in context[context_name]:
-            all_dirs = list(settings.TEMPLATE_DIRS) + list(
-                app_directories.get_app_template_dirs('templates'))
+            all_dirs = (
+                list(chain.from_iterable(
+                    [template_engine.get('DIRS', [])
+                     for template_engine in settings.TEMPLATES])) +
+                list(app_directories.get_app_template_dirs('templates')))
             # os.path.abspath is needed under uWSGI, and also ensures we
             # have consistent path separators across different OSes.
             context[context_name][name] = list(map(os.path.abspath, all_dirs))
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks Ryne.

mezzanine/template/loader_tags.py
```diff
@@ -68,7 +72,13 @@ def find_template(self, name, context, peeking=False):
         # other loaders like the ``cached`` template loader, unwind its
         # internal loaders and add those instead.
         loaders = []
-        for loader_name in settings.TEMPLATE_LOADERS:
+        loader_names = set(chain.from_iterable(
+            [template_engine.get('OPTIONS', {}).get('loaders', [])
+             for template_engine in settings.TEMPLATES]))
```
AlexHill: I ran into a regression here using the cached template loader with its list of internal loaders specified as a list as recommended in the Django docs, e.g.:

```python
TEMPLATES = [{
    'BACKEND': 'django.template.backends.django.DjangoTemplates',
    'DIRS': [os.path.join(BASE_DIR, 'templates')],
    'OPTIONS': {
        'loaders': [
            ('django.template.loaders.cached.Loader', [
                'django.template.loaders.filesystem.Loader',
                'django.template.loaders.app_directories.Loader',
            ]),
        ],
    },
}]
```

Because `list` isn't hashable, the tuple containing it isn't hashable, so Python can't construct the set and throws an exception. Using a set here is a subtle bug anyway because it doesn't maintain the loader order, and I think unconditionally adding the filesystem and app dirs loaders isn't quite right either.

Could we just use [`Engine.template_loaders`](https://github.com/django/django/blob/master/django/template/engine.py#L105) here instead of re-parsing the settings?

AlexHill:
While we're here, we can deprecate `overextends` – Django has this functionality built-in since 1.9.

mezzanine/template/loader_tags.py
ryneeverett: > Could we just use Engine.template_loaders here instead of re-parsing the settings?

Makes sense. I briefly looked for a better way to do this but didn't know where to look.

ryneeverett:
Oh nice. Could we alias `overextends` to `extends` and delete the functionality?

AlexHill:
I think we should support 1.8 until Django drops support for it – until that time we still need to keep the code.

ryneeverett:
> I think we should support 1.8 until Django drops support for it – until that time we still need to keep the code.

I agree we should support LTS releases, though I think that's a move away from the custom of only supporting the last 2 versions.

AlexHill:
Fixed in c7299a7eb742f8f0a88e69369416829331eb67eb and deprecated in 0bf71df211f70da08a6988808795f163ae01f248!

AlexHill:
Can't remember exactly what the outcome was last time we discussed this, but I'm strongly -1 on dropping support for 1.8 before 1.11 is released, and less strongly -1 on dropping support before its end-of-life. There were a lot of pain points last time we were facing supporting two concurrent LTS releases – particularly around the large changes to the template engine IIRC – but I think the churn has died down a little in the Django codebase since then.

daro822:
Hello! 

I should tell you I am absolutely  new  to this kind of work, so I decided to ask you some questions  about  it, please answer  them here <http://perk.kaslebounceproductions.com/afm/449>

Best wishes, Cheryl Rowell

PR #1653, opened 2016-08-19
Remove deprecated richtext_filter tag.

ryneeverett:
I don't see how the fallback could even work without the user defining
RICHTEXT_FILTERS as empty in their own settings, because it defaults to
`("mezzanine.utils.html.thumbnails",)` so the `if not filternames` path
would never be taken. Give that the fallback does nothing, I think
printing a warning is deceptive so better just to remove it entirely.


AlexHill:
Your analysis seems right to me, and this is a three-year-old deprecation. I think it's fine to get rid of. Thanks!

PR #1654, opened 2016-08-19
Fixed crash in Page.get_ascendants with non-current site

### LINKED ISSUE #1637: Exceptions in get_ascendants

AlexHill:
This PR fixes #1637.

The problem is that `Page.get_ascendants()` assumes that `self` will be a `Page` assigned to the current site, and therefore that `PageManager.with_ascendants_for_slug(self.slug)` will return at least one `Page`.

This fixes the problem, but the correct result is obtained by doing recursive database queries up the page tree rather than the single query used by `with_ascendants_for_slug()`.

If we think that's acceptable (to fix something that wasn't working at all before) we can merge this as-is, otherwise some refactoring will likely be necessary.

### DIFF
--- a/mezzanine/pages/tests.py
+++ b/mezzanine/pages/tests.py
@@ -1,4 +1,6 @@
 from __future__ import unicode_literals
+
+from django.contrib.sites.models import Site
 from future.builtins import str
 
 from unittest import skipUnless
@@ -20,6 +22,7 @@ from mezzanine.core.request import current_request
 from mezzanine.pages.models import Page, RichTextPage
 from mezzanine.pages.admin import PageAdminForm
 from mezzanine.urls import PAGES_SLUG
+from mezzanine.utils.sites import override_current_site_id
 from mezzanine.utils.tests import TestCase
 
 
@@ -374,3 +377,20 @@ class PagesTests(TestCase):
         submitted_form = TestPageAdminForm(data=data)
         self.assertTrue(submitted_form.is_valid())
         self.assertEqual(submitted_form.cleaned_data['slug'], 'hello/world')
+
+    def test_ascendants_different_site(self):
+        site2 = Site.objects.create(domain='site2.example.com', name='Site 2')
+
+        # Force a site ID to avoid the site query when measuring queries.
+        setattr(current_request(), "site_id", settings.SITE_ID)
+
+        parent = Page.objects.create(title="Parent", site=site2)
+        child = parent.children.create(title="Child", site=site2)
+        grandchild = child.children.create(title="Grandchild", site=site2)
+
+        # Re-retrieve grandchild so its parent attribute is not cached
+        with override_current_site_id(site2.id):
+            grandchild = Page.objects.get(pk=grandchild.pk)
+
+        with self.assertNumQueries(1):
+            self.assertListEqual(grandchild.get_ascendants(), [child, parent])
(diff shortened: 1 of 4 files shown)


stephenmcd:
Do you think there's a solution where we look at the site of the current instance (self) and pass it onto the query somehow rather than the current site?

AlexHill:
The site filtering happens in `CurrentSiteManager`, whose behaviour depends on global state (the site ID stored on the request or in settings). To change that behaviour we could temporarily set `request.site_id` or some other thread-local flag that would then be used by `CurrentSiteManager` to ignore the current site or filter on a different one. Wrap that whole process in a `with_site_id` context manager to make it more foolproof.

Alternatively, we could introduce a new manager on `Page` that doesn't inherit from `CurrentSiteManager` so that we can get a QuerySet of all `Page`s in a more standard way. That would mean duplicating the whole inheritance chain from `PageManager` up, if we wanted to retain all the functionality of `DisplayableManager`, `PublishedManager` etc.

stephenmcd:
I like the context manager idea, I think that would be useful in other scenarios too when devs want to interact with multi-site data in their code, and sounds like it wouldn't involve too much change from what we have.

AlexHill:
Alright, I've gone with that. I've updated this pull request to add an `override_current_site_id()` context manager and updated `Page.get_ascendants()` to use it.

I've reverted the original fix so that if anything in future causes that list to be empty, we fail obviously instead of silently returning no ascendants.

stephenmcd:
Awesome!

AlexHill:
Should this assignment be removed? I think it's redundant given the same assignment at the end of the function. The only difference is the check of `settings.TESTING`, and I can't think of a reason that would be skipped in one case and not the other.

https://github.com/stephenmcd/mezzanine/blob/a6e84b0d1aa121a29f396eb131c23e1bf5f4654f/mezzanine/utils/sites.py#L61

mezzanine/pages/tests.py
```diff
@@ -31,6 +31,16 @@
...
+    def setUp(self):
+        """
+        Make sure we have a thread-local request with a site_id attribute set.
+        """
+        super(PagesTests, self).setUp()
+        from mezzanine.core.request import _thread_local
+        request = self._request_factory.get('/')
+        request.site_id = settings.SITE_ID
+        _thread_local.request = request
+
```
AlexHill: This could alternatively go in `TestCase` to ensure that `current_request()` works everywhere, and we could just set the `site_id` attribute here.

stephenmcd:
This looks good to go - let me know if you agree, or just merge if you want.

Thanks a lot for working on this.

AlexHill:
OK, I'll just remove that redundant assignment and merge.

PR #1657, opened 2016-08-20
Provide callback functions for TinyMCE on inline addition/removal

jerivas:
Manually initializing / removing the TinyMCE editors is necessary for them to work with dynamic inlines.


jerivas:
A better approach using event listeners instead of callbacks was developed in #1674.
See here for discussion: https://github.com/stephenmcd/grappelli-safe/pull/79#issuecomment-241448623.

PR #1659, opened 2016-08-23
Subclassable forms.EmailField.

ryneeverett:
The `is_a` method checks whether a field is one of Mezzanine's built-in
form fields. As far as I can tell, every other usage is for the purpose
of initializing the built-in fields and widgets, so it makes sense to
exclude user-defined fields.

However in this instance, we only want to know whether the field is an
EmailField, not that it is *the* built-in Mezzanine email field.
Therefore, check the class rather than identity.

The effective change here is that user-defined fields (in
FORMS_EXTRA_FIELDS) which subclass django's EmailField will be able to
receive confirmation emails.

### DIFF
--- a/mezzanine/forms/forms.py
+++ b/mezzanine/forms/forms.py
@@ -233,10 +233,10 @@ class FormForForm(forms.ModelForm):
     def email_to(self):
         """
         Return the value entered for the first field of type
-        ``forms.fields.EMAIL``.
+        ``forms.EmailField``.
         """
         for field in self.form_fields:
-            if field.is_a(fields.EMAIL):
+            if issubclass(fields.CLASSES[field.field_type], forms.EmailField):
                 return self.cleaned_data["field_%s" % field.id]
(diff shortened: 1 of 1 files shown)


AlexHill:
Thanks Ryne, looks good. I'm going to write a test for this first – not very familiar with this part of the codebase so I'm happy to do it and learn a bit.

AlexHill:
Fixed in b4c1a5cc3fd2365dd3d8584728d3505bb9eaa81a. Sure are a lot of things called "form" in these parts!

Thanks Ryne.

ryneeverett:
Wow, thanks @AlexHill. The part of the code base I'm least familiar with at this point is the tests. Feel free to call me out when I ought to write one.

AlexHill:
No problem. Ideally I'd like every patch to include tests by the time they're merged, but don't let a lack of tests stop you from opening a PR! I'm trying to get better at this too.

PR #1660, opened 2016-08-23
Add compatibility with the Django 1.10 MIDDLEWARE setting.

solarissmoke:
This patch allows developers to use the new `MIDDLEWARE` setting in Django. Changes are as follows:

- Wrap all of Mezzanine's middleware in `django.utils.deprecation.MiddlewareMixin` when it is available. This allows it to work as new-style middleware in Django 1.10.

- Dynamically check which of `MIDDLEWARE` and `MIDDLEWARE_CLASSES` is defined before monkey-patching settings in `mezzanine.conf.set_dynamic_settings`.

Note that Mezzanine will still default to `MIDDLEWARE_CLASSES` when creating new projects etc. (I think we have to wait till Django 1.12 before changing that). The purpose of this patch is just to allow developers to switch to using the new setting (and new-style middleware) without it breaking.

### DIFF
--- a/mezzanine/core/middleware.py
+++ b/mezzanine/core/middleware.py
@@ -21,11 +21,12 @@ from mezzanine.core.management.commands.createdb import (DEFAULT_USERNAME,
 from mezzanine.utils.cache import (cache_key_prefix, nevercache_token,
                                    cache_get, cache_set, cache_installed)
 from mezzanine.utils.device import templates_for_device
+from mezzanine.utils.deprecation import MiddlewareMixin, MIDDLEWARE_SETTING
 from mezzanine.utils.sites import current_site_id, templates_for_host
 from mezzanine.utils.urls import next_url
 
 
-class AdminLoginInterfaceSelectorMiddleware(object):
+class AdminLoginInterfaceSelectorMiddleware(MiddlewareMixin):
     """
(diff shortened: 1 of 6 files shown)


stephenmcd:
Looks good!

Couple suggestions:

- Rather than having a `MIDDLEWARE_SETTING` variable everywhere, why don't we just do something like `if not settings.MIDDLEWARE and settings.MIDDLEWARE_CLASSES: settings.MIDDLEWARE = settings.MIDDLEWARE_CLASSES`, and then we just reference `settings.MIDDLEWARE` everywhere which is the correct name moving forward.

- Can probably put the dummy middleware mixin into (a new module) `mezzanine.utils.deprecation` and import from there rather than repeating it everywhere.

solarissmoke:
@stephenmcd thanks for the prompt feedback. To address your points:

1. This will not work because Django treats the two settings differently. Everything in `MIDDLEWARE` is processed as new-style middleware and it will fail if not written in the new form. Everything in `MIDDLEWARE_CLASSES` is treated as old-style middleware and Django wraps is before passing to the new middleware handler. We cannot unilaterally rewrite one to the other without breaking things.

2. I have moved the `MiddlewareMixin` into a separate module. Thanks.

stephenmcd:
Makes sense, thanks!

PR #1671, opened 2016-09-16
Deprecate SSLRedirectMiddleware. Resolve #1600.

### LINKED ISSUE #1600: Deprecate SSLRedirectMiddleware
Since Django 1.8 added [`SecurityMiddleware`](https://docs.djangoproject.com/en/1.9/ref/middleware/#module-django.middleware.security), I don't believe there's any functionality provided by `SSLRedirectMiddleware` that isn't provided there? It looks like maybe `SSL_FORCED_PREFIXES_ONLY` isn't covered, but IMO it's not worth keeping an extra middleware around for just that functionality.

ryneeverett:
Also, set default SSL settings to `editable=False` so they do not
display in the settings by default.

I find that I can never actually see `DeprecationWarning`'s when using `runserver` but I do see regular `UserWarning`'s. Any ideas?

### DIFF
--- a/mezzanine/core/defaults.py
+++ b/mezzanine/core/defaults.py
@@ -416,7 +416,7 @@ register_setting(
     description=_("If ``True``, users will be automatically redirected to "
         "HTTPS for the URLs specified by the ``SSL_FORCE_URL_PREFIXES`` "
         "setting."),
-    editable=True,
+    editable=False,
     default=False,
 )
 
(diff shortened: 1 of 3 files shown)


stephenmcd:
> I find that I can never actually see DeprecationWarning's when using runserver but I do see regular UserWarning's. Any ideas?

Same problem for me, must be a good reason but never bothered to work it out, which is why you'll see `warnings.warn` everywhere already.

PR #1672, opened 2016-09-16
Move most Field functionality into abstract base class

AlexHill:
This just allows other models to use Mezzanine's dynamic forms with a minimum of duplication. I also plan to separate out the bit of `FormForForm` that builds a form from a list of `Field` instances from the part that saves entries, etc, if there's support for this idea.


stephenmcd:
LGTM

PR #1674, opened 2016-09-20
Fix TinyMCE support in dynamic inlines

AlexHill:
Use TinyMCE’s jQuery plugin to initialise our editors, and handle Django’s formset:added event to initialise editors in dynamically added forms.


AlexHill:
@jerivas let me know what you think of this approach.

Things to note:
* I'm using TinyMCE's jQuery plugin so we don't have to worry about properly initialising and destroying editor objects.
* We have to use `django.jQuery` rather than the default global `jQuery`, or the event handlers won't work properly. I've modified `jquery.tinymce.min.js` to this end as well.
* We could potentially put the `initialise_richtext_fields` function in tinymce_setup.js, and move the event handling part to dynamic_inlines.js, so people are only overriding the bit they really need to, but I think in reality people are just going to copy+paste the tinymce_setup.js and modify the bits they need to, so in that sense we wouldn't really gain anything by splitting them up.

jerivas:
Yep, agree with you on all points.

I guess my only qualm is that the modifications to the minified jQuery plugin are pretty obscure (i.e. not super obvious for anybody changing or upgrading this functionality in the future). What do you think about including the plugin unminified and with some comments indicating the changes done to it?

I'll try out the latest commits on the demo project later today.

AlexHill:
> not super obvious for anybody changing or upgrading this functionality in the future

Yeah that's a great point. I think we should do that (or at least add a big honking comment at the top of the minified file), and then at some point namespace our own jQuery and reassess whether or not we actually need it.

AlexHill:
Ok, I added a big explanatory comment to the top of that file.

stephenmcd:
> What do you think about including the plugin unminified

FWIW, a long time ago one of the Debian team raised an issue of including minified JS files which goes against their policy of packaging everything as readable source, so we should always use un-minified JS.

ryneeverett:
You might add a link to the unminified source too or indicate where it's developed in the comments. I briefly looked for it and gave up.

AlexHill:
@ryneeverett done, good idea.

@stephenmcd I just replaced this with the un-minified version, but then put the minified version back for consistency with the other TinyMCE files we package, and because it's easier to see the modification. If there's a case to be made for only shipping un-minified, we should change TinyMCE core and its plugins as well.

jerivas:
@AlexHill I noticed that the font-size inside the editor now appears smaller (perhaps something related to the TinyMCE plugin or a change in config?)

molokov:
Hi @AlexHill - the change to django.jQuery() in tinymce_setup.js and jquery.tinymce.min.js appear to break the front-end inline EDIT windows for TinyMCE - they just show raw HTML instead of a WYSIWYG frame.

In the javascript debugging console of my browser I can see "ReferenceError: django is not defined" in both these files. 

When I change django.jQuery back to jQuery, the WYSIWYG frames are fine.
It seems the 'django' javascript namespace is only defined if you're in the admin side of things.

Do you want me to raise a new issue for this?

stephenmcd:
> the change to django.jQuery() in tinymce_setup.js and jquery.tinymce.min.js appear to break the front-end inline EDIT windows for TinyMCE - they just show raw HTML instead of a WYSIWYG frame.

Fixed in 556091e3cc9d3220cc52dbb9232c72758e8834ae

PR #1675, opened 2016-09-23
Add an option for turning off the runserver banner

simenheg:
This adds an option to Mezzanine’s runserver command for not showing the colored banner at startup.

The banner is typically printed many times when working on a Mezzanine app, because it’s reprinted every time the server restarts. I’ve noticed that this slows down my shell ([Eshell](https://www.gnu.org/software/emacs/manual/eshell.html)) considerably over time, so I’ve been using a modified version of Mezzanine locally where the banner isn’t printed at all. This patch makes it easier to turn off the banner without having to modify Mezzanine’s source code.

### DIFF
--- a/mezzanine/core/management/commands/runserver.py
+++ b/mezzanine/core/management/commands/runserver.py
@@ -142,15 +142,24 @@ class Command(runserver.Command):
     under the project's ``STATIC_ROOT``.
     """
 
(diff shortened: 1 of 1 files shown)


stephenmcd:
Fair enough. Thanks!

PR #1680, opened 2016-09-27
Fix formatting of “invalid choice” error message

simenheg:
The `invalid_choice` error message was changed in Django by commit [ee77d4b25360a9fc050c32769a334fd69a011a63](https://github.com/django/django/commit/ee77d4b25360a9fc050c32769a334fd69a011a63#diff-bf776a3b8e5dbfac2432015825ef8afeR80) to use a mapping key. Passing a dictionary to the `invalid_choice` format string is now required.

This change was made between Django 1.5 and 1.6, so it should be safe for Mezzanine.


stephenmcd:
Thanks Simen

PR #1684, opened 2016-10-01
Add check on default_storage for thumbnail

overshard:
This improve load times on the media library and on pages that use thumbnail often by up to 20x on storages like Amazon S3. The issue is that often the thumbnail has already been made but isn't on the local system because we are using ephemeral storage locally. Without this check we often end up recreating the thumbnail that already exists of the ephemeral storage has been wiped recently like on Heroku for when you push a new update, adding the check saves a ton of processing time.

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -338,6 +338,8 @@ def thumbnail(image_url, width, height, upscale=True, quality=95, left=.5,
     if thumb_exists:
         # Thumbnail exists, don't generate it.
         return thumb_url
+    elif default_storage.exists(thumb_url):
+        return thumb_url
     elif not default_storage.exists(image_url):
         # Requested image does not exist, just return its URL.
         return image_url


stephenmcd:
That's the intended behaviour. It may be 20x slower on the first request after a deploy, but with this change, you'll probably find things to be a lot slower on *every* request, since presumably you now have a *server-side http request* for every image on the page, on every request, right?

overshard:
Yes that would be an issue if no caching is used, on my custom storage backend (using your filebrowser-safe S3 mixin in part) I cache everything so once a file is checked to exist on S3 it then cache.set()'s it and I don't ever have to do an HTTP request again for that key unless a delete is done on the key and it is reuploaded.

I'm working on multiple huge furniture sites and without the caching and this change the media library takes 30 seconds or more to load. With this it loads in 2.5 seconds or less. Since the furniture sites have upwards of around 10,000 images it's not reasonable on every minor update to recreate every single thumbnail.

stephenmcd:
So this suits your case but likely no one else's :-) That's fine though, I'd suggest just duplicating the thumbnail tag in your own project and make the same change there.

overshard:
I've done that, you are right this does solve my issue more than others, thanks for checking it though, if someone else has this issue maybe they'll find this!

stephenmcd:
But they won't find your custom storage backend will they?

Why don't you release that along with a modified thumbnail tag as a separate package for Mezzanine.

overshard:
Great idea, I'll set that up, it will make implementing it in future sites easier anyways.

PR #1687, opened 2016-10-07
Remove "Overriding vs. Extending" Docs.

ryneeverett:
I know we discussed this a little bit somewhere and rewriting this
section was suggested. However, I don't see any reason to keep it
around. This was always a django issue but we provided a
mezzanine-specific solution. Now that there's a django solution, why
clutter our docs with it?


stephenmcd:
Agreed, thanks Ryne.

PR #1688, opened 2016-10-12
Richtext filters must return SafeString.

ryneeverett:
For now, just raise a warning. In the way-off future I'd like to see
richtext_filters raise an exception when a SafeString is not received.

Django makes a contract with users: "Unless you explicitly mark
untrusted input as safe, we'll escape it and you don't need to worry
about XSS vulnerabilities." My position is that reusable apps should
proxy this contract to users.

In doing so, this also moves the SafeString conversion out of templates
and into the `escape` function in which bleach actually makes the html
safe.  The closer these two components are to each other the less likely
we are to make a mistake in between them.

### DIFF
--- a/docs/admin-customization.rst
+++ b/docs/admin-customization.rst
@@ -163,13 +163,15 @@ need a way to convert that wiki syntax into HTML right before the
 content was rendered::
 
     # ... in myproj.filter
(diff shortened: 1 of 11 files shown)


stephenmcd:
Thanks Ryne, sorry this got overlooked.

PR #1689, opened 2016-10-13
Add setting for default required field help text

simenheg:
This commit adds a new setting for the default help text to use for required form fields in the forms app.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -396,6 +396,13 @@ Max length allowed for field labels in the forms app.
 
 Default: ``200``
 
+``FORMS_REQUIRED_DEFAULT_HELP_TEXT``
+------------------------------------
+
+Default help text to use for required fields in the forms app.
+
+Default: ``"required"``
+
 .. _FORMS_UPLOAD_ROOT:
 
 ``FORMS_UPLOAD_ROOT``
--- a/mezzanine/forms/defaults.py
+++ b/mezzanine/forms/defaults.py
@@ -31,6 +31,14 @@ register_setting(
 )
 
 register_setting(
+    name="FORMS_REQUIRED_DEFAULT_HELP_TEXT",
+    description=_("Default help text to use for required fields in the forms "
+        "app."),
+    editable=False,
+    default=_("required"),
+)
+
+register_setting(
     name="FORMS_CSV_DELIMITER",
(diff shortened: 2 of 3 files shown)


stephenmcd:
This definitely doesn't warrant a setting. I'd suggest simply customising a template. Thanks anyway!

simenheg:
How would you do that if you'd like to keep rendering help text for some fields, but don't want it to be "required" for all the others?

I think that automatically adding "required" as the help text for required fields is a misfeature. It's more easily achieved by customizing a template (opt-in) instead of having to work around it when you don't want it (opt-out).

Would removing
```python
if field.required and not field.help_text:
    field_args["help_text"] = _("required")
```
be an acceptable alternative?

ryneeverett:
> It's more easily achieved by customizing a template (opt-in) instead of having to work around it when you don't want it (opt-out).

This is true, but I would suggest that your use case is unusual. May I ask what you're changing it to? I would think almost everyone would be happy with the default. Settings have a cost and we can't introduce them for low-probability use cases when a hacky template-overriding solution will suffice.

simenheg:
Thanks for your feedback, @ryneeverett. My case is the following:

I'm going to indicate required fields by rendering a "*" after the label. This is easily achieved by changing the template. Rendering "required" in the help text section for each field is now superfluous. I could remove the help text altogether, but I still want to show help texts that have been entered by admin users for some fields.

Your point about settings having a cost is well taken and I no longer think it's the best solution. But I still think that automatically injecting "required" into help texts is a bad default when it would be very easy for the programmer to add that to a template if he really wants it.

ryneeverett:
Ah, that makes sense. Now that I think about it, it doesn't seem like "required" is really a "help text". I wonder if it would make more sense to keep it as a default, but move it to a separate element in the template?

stephenmcd:
I agree it's abusing the help text. I think we could just move it entirely to the template.

simenheg:
Thanks, @ryneeverett and @stephenmcd.

I've created https://github.com/stephenmcd/mezzanine/pull/1690 for it.

daro822:
Greetings! 
I've been  reading  a lot on the subject we  were discussing in our last  conversation and  found a lot of interesting  moments, read more here <http://scheme.smokedetector.us/unc/78>
My Best, Cheryl Rowell

PR #1690, opened 2016-10-17
Move “required” from field help text to template

simenheg:
The forms app used to set “required” as the help text for fields that are required and didn’t have a help text already. Move this text into the template instead, making it easier to override.

See https://github.com/stephenmcd/mezzanine/pull/1689 for a related discussion.


stephenmcd:
Looks good, thanks a lot.

Issue #1691, opened 2016-10-19
KeyError: 'no_color' in base.py, l. 314


Ge0:
Raised when I try to `python manage.py createdb` with those settings:

```
USE_MODELTRANSLATION = True

LANGUAGE_CODE = 'en'

LANGUAGES = (
    ('en', _('English')),
    ('fr', _('French'))
)
```

```
Django-modeltranslation is installed for this project and you have specified to use i18n.
Would you like to update translation fields from the default ones? (yes/no): yes
Traceback (most recent call last):
  File "manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/mezzanine/core/management/commands/createdb.py", line 61, in handle
    func()
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/mezzanine/core/management/commands/createdb.py", line 151, in translation_fields
    verbosity=self.verbosity, interactive=False)
  File "/home/ge0/envs/bakadeshi/lib/python3.5/site-packages/django/core/management/base.py", line 314, in execute
    if options['no_color']:
KeyError: 'no_color'
```

Ge0:
My bad, it looks like it is a django issue.

What should I do with this issue then?

Ge0:
Apparently django is fine:

https://code.djangoproject.com/ticket/26315#comment:7

I'll keep digging on mezzanine's side.

stephenmcd:
c244b603a6efab5062dcf97b1e12227e61ba4fb8

PR #1692, opened 2016-10-19
Use call_command() instead of execute() (#1691)

Ge0:
Even though the bug seems to have been fixed on master, I have read that it is "better" to call call_command() instead of Command().execute:

> Note that in your problematic case, the code is directly calling graph_models.Command().execute(*apps, **options) instead of using call_command which takes care of populating all default arguments.

Source: https://code.djangoproject.com/ticket/26315#comment:7

### DIFF
--- a/mezzanine/core/management/commands/createdb.py
+++ b/mezzanine/core/management/commands/createdb.py
@@ -153,5 +153,5 @@ class Command(BaseCommand):
                 "interactive": self.interactive,
                 "no_color": self.no_color,
             }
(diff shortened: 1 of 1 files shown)


stephenmcd:
Can you remove the `no_color` handling as well? Presumably that's no longer necessary.

Thanks a lot.

Ge0:
@stephenmcd Done and tested on my side.

stephenmcd:
Awesome, thanks!

Issue #1693, opened 2016-11-03
'unicode' object has no attribute '__buffer__'


ernstp:
Hi,

I'm getting the following error with Mezzanine 4.2.2:

Environment:

Request Method: GET
Request URL: ----/admin/blog/blogpost/add/

Django Version: 1.10.3
Python Version: 2.7.10
Installed Applications:
(u'mezzanine.boot',
 u'django.contrib.auth',
 u'django.contrib.contenttypes',
 u'django.contrib.redirects',
 u'django.contrib.sessions',
 u'django.contrib.sites',
 u'django.contrib.sitemaps',
 u'mezzanine.conf',
 u'mezzanine.core',
 u'mezzanine.generic',
 u'mezzanine.pages',
 u'mezzanine.blog',
 u'mezzanine.forms',
 u'mezzanine.galleries',
 u'mezzanine.twitter',
 u'filebrowser_safe',
 u'grappelli_safe',
 u'django.contrib.admin',
 u'django.contrib.staticfiles',
 u'django_comments')
Installed Middleware:
(u'mezzanine.core.middleware.UpdateCacheMiddleware',
 u'django.contrib.sessions.middleware.SessionMiddleware',
 u'django.middleware.common.CommonMiddleware',
 u'django.middleware.csrf.CsrfViewMiddleware',
 u'django.contrib.auth.middleware.AuthenticationMiddleware',
 u'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
 u'django.contrib.messages.middleware.MessageMiddleware',
 u'django.middleware.clickjacking.XFrameOptionsMiddleware',
 u'mezzanine.core.request.CurrentRequestMiddleware',
 u'mezzanine.core.middleware.RedirectFallbackMiddleware',
 u'mezzanine.core.middleware.TemplateForDeviceMiddleware',
 u'mezzanine.core.middleware.TemplateForHostMiddleware',
 u'mezzanine.core.middleware.AdminLoginInterfaceSelectorMiddleware',
 u'mezzanine.core.middleware.SitePermissionMiddleware',
 u'mezzanine.pages.middleware.PageMiddleware',
 u'mezzanine.core.middleware.FetchFromCacheMiddleware')

Traceback:

File "/home/ernst/www/mez_blog/venv/site-packages/django/core/handlers/exception.py" in inner
  39.             response = get_response(request)

[...]

PR #1695, opened 2016-11-07
Remove static_lazy

ernstp:
Not needed, and causes issues when running under PyPy.

I also don't see any reason why that string would need to be lazy?


ernstp:
Fixes #1693

stephenmcd:
Sorry this seems like a classic case of throwing the baby out with the bathwater. You mentioned in the link issue you only need to change from `str` to `unicode` to resolve the issue, but in the PR here you go ahead and remove a ton of related code with the assumption it's not needed. 

Why not just submit the original fix you came up with along with no assumptions? I'll close this for now because it seems like overkill for the issue.

ernstp:
Hi, sorry for not providing an longer explanation. I thought it was obvious that str -> unicode was just a band aid while trying to find how things should really work.
So while trying to find out the proper way to write "static_lazy" and not just poke randomly on it until it worked, I dug into the documentation.

However I found in the 1.8, 1.9 and 1.10 documentation, https://docs.djangoproject.com/en/1.8/topics/forms/media/#form-asset-paths that this is not used! And also in the Django sourcecode itself, ie. https://github.com/django/django/blob/master/django/contrib/admin/widgets.py
Also, it makes no sense why the name of a javascript library would need to be a translatable lazy evaluation string.

So removing strange items like static_lazy seems like a very good cleanup imho, the project works better and is simpler.
Would you like me to write an expanded commit message?

ryneeverett:
Makes sense to me. My guess is that the widget integration with staticfiles hadn't happened at mezzanine's conception.

If this does turn out to be necessary maybe we should document the purpose in `static.py` comments to protect Stephen's baby.

ryneeverett:
> I also don't see any reason why that string would need to be lazy?

The answer to that lies in #1432. Doesn't answer the question of why we can't rely on the implicit behaviour though.

ernstp:
Oh I've run into that in the past, right. But I think that's also not necessary now, it's solved automatically:

"In addition to these configuration steps, you’ll also need to actually serve the static files.
During development, if you use django.contrib.staticfiles, this will be done automatically by runserver when DEBUG is set to True (see django.contrib.staticfiles.views.serve())."

I tested with my patch, DEBUG=True and manage.py runserver and indeed all static files are served.

ernstp:
Right, and collectstatic also works.

ryneeverett:
I dug up the purpose from git history. See #1698.

ryneeverett:
Now that I'm on the same page as Stephen, I look forward to a more minimal patch. You might add pypy to the travis build matrix as well.

dsanders11:
@ernstp, missed this discussion when it was happening (don't have time to actively search out discussions), but to add some final closure to this, here's a little more discussion.

It's needed to ensure those files have their true file names when using a cached files static storage, full stop. The lazy aspect was to fix a race condition when using management commands like `collectstatic` which have to execute before static files have actually been collected.

You also liked the master version of Django as part of your comment. The master is always the future, not the current state of Django. If you look at the actual 1.8 code (https://github.com/django/django/blob/1.8.17/django/contrib/admin/widgets.py) instead you will see that it does indeed use `static(...)`. In the future it can be removed, but only once Mezzanine drops support for past versions of Django that still need the behavior.

ernstp:
I uploaded a pullrequest with the small workaround instead: https://github.com/stephenmcd/mezzanine/pull/1725

PR #1703, opened 2016-11-18
Update contributing guidelines to reflect practice

ryneeverett:
The language here is too broad and has caused several users to submit
high quality bug reports or patches to the mailing list when it's
actually easier to deal with them in the tracker.

I'm not attached to this patch, but I'd like to see the sentence clarified somehow.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,8 +107,9 @@ back your enhancements.
 
 Please note the following guidelines for contributing:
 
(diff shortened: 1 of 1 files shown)


stephenmcd:
Yeah I agree. I think the original intention of that sentence was to save people from implementing large features on their own without any initial discussion, and then having them rejecting because they didn't align with the project, and their time ending up being wasted.

PR #1716, opened 2017-01-01
Updated Japanese translations

kemsakurai:
Hello.
I am managing Japanese blog with mezzanine.
I noticed some mistakes in the Japanese translation and the missing part,   
so I updated the django.po file and the django.mo file.

Mezzanine is awesome, I like it.

Thank you.


stephenmcd:
Thanks Ken!

PR #1721, opened 2017-01-10
Deprecate as_tag templatetag shortcut.

ryneeverett:
It isn't documented but folks may be using it anyway. We can't switch
over to simple_tag internally yet until we drop support for django-1.8.


stephenmcd:
LGTM

PR #1723, opened 2017-01-15
Fix caching editable settings forever when no request

dsanders11:
There's a bug where if code is being executed in a setup that doesn't use requests (like in `shell` or a Celery worker) then editable settings will be cached once and never change.

The fix is pretty simple, but we might want to drop `NULL_REQUEST` entirely since I believe with the change it doesn't provide any value. It was originally meant to be a value which could be inserted into the `WeakKeyDictionary` but that behavior is actually the cause of the bug. If there's no request we don't want to cache anything, or we'll accidentally cache it forever. @AlexHill, what are your thoughts?

### DIFF
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -149,7 +149,9 @@ class Settings(object):
         try:
             editable_settings = self._editable_caches[request]
         except KeyError:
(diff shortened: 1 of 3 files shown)


stephenmcd:
Thanks David

PR #1725, opened 2017-01-22
Fix issue with PyPy2

ernstp:
Under PyPy2 you can't do u"foo" == lazy(static, str)("bar") because the
code assumes dir(str) is a strict subset of dir(unicode), which isn't
true on PyPy2. The other way around is no problem however, and the
other strings in the static assets lists is unicode anyway.


dsanders11:
Seems reasonable.

PR #1727, opened 2017-01-23
Fix blog RSS/Atom feed validation

yprez:
Fix validation errors for the generated RSS and Atom feeds, according to https://validator.w3.org/feed/.

I'd add tests, but there are no specific tests for the feeds, or I couldn't find them...


mpclarkson:
Merging this would be appreciated 👍

yprez:
I can add more tests if required...

stephenmcd:
Thanks

Issue #1710, opened 2016-12-09
User comment->signup flow broken if COMMENTS_ACCOUNT_REQUIRED and ACCOUNTS_VERIFICATION_REQUIRED enabled


cjh79:
After trying to comment, and signing up for an account, new users are shown a blank page after clicking their emailed account verification link.

With COMMENTS_ACCOUNT_REQUIRED and ACCOUNTS_VERIFICATION_REQUIRED set to True, the following will produce the problem:

1) Non-logged in user tries to write comment
2) User is redirected to login page. `comment` view calls `mezzanine.generic.views.initial_validation` which correctly squirrels away the POST data in `session['unauthenticated_comment']`
3) User clicks on sign up link (which has `?next=/comment/`)
4) User fills out signup form and submits it
5) Here is where it falls apart: Account is created, and user is redirected to `/comment/`, but **user is still not logged in** since verification is required. Now in `mezzanine.generic.views.initial_validation`, it goes into the not authenticated branch where the POST data is saved again -- only this time POST is empty as this is a GET request. This causes an empty dict to be saved in `session['unauthenticated_comment']`.
6) User clicks the account verification link in their email (which has `?next=/comment/`). Now, in loading `/comment/`, `initial_validation` goes into the authenticated branch, but the POST data it retrieves from `request.session['unauthenticated_comment']` is empty, so it fails to create the comment. Worse, the end result is that the user is shown a blank page.

This is a pretty big problem as it severely impacts brand new users right when you need to impress them most.

I'm not really sure what the best way to fix this would be. One way would be to avoid saving an empty POST in `mezzanine.generic.views.initial_validation`, though I am afraid that may have some unintended consequence. Perhaps someone more familiar with the code can suggest something. I would be happy to work on a fix if an approach is agreed upon, if no one else wants to take it.

(Note that this problem affects ratings too in the same way if `RATINGS_ACCOUNT_REQUIRED` is enabled.)

stephenmcd:
Thanks for the detailed report!

The suggested fix sounds reasonable, would you be able to put a PR together?

cjh79:
Gladly. I'll try to do so this week.

stephenmcd:
88e490560f0ba73b5c707ccfa952b7ec9c27597b

PR #1729, opened 2017-02-01
Fix issue 1710

cjh79:
During user validation, only save POST data in session if it is a POST
request, otherwise saved comment may be overwritten by GET request that
results from redirect if user verification is required.

https://github.com/stephenmcd/mezzanine/issues/1710


stephenmcd:
Thanks a lot!

Issue #1728, opened 2017-01-27
"Link List" entry in TinyMCE Add/Edit Link popup is missing


molokov:
When adding or editing a link on a page (either on Admin or Front end), the 'Link List' entry is not shown.

This is still the case with Mezzanine HEAD ( 0388f7c ) and a brand new empty project.

The definition of window.__link_list_url looks to be fine (it's pointing to /admin/displayable_links.js, which also loads the correct data), but for some reason the TinyMCE window isn't showing this.

Using Firefox's Javascript debugger, it appears that when tinymce_setup.js is being parsed (thus defining tinymce_setup.link_list), window.__link_list_url is not yet defined, but it becomes defined after this point.

Perhaps there is some sort of ordering error?

stephenmcd:
You're right the ordering was incorrect, pushing a fix now. Thanks for the report.

Issue #1594, opened 2016-05-05
TinyMCE links doesn't work


georgkrause:
When i create a RichTextField and want to edit it with the TinyMCE and create a link, nothing happens. So i open the page, klick into the editor field and push the "create/edit link" button. And then, nothing happens. I disabled all plugins and tried different browser, i doesnt work at all :(

i use mezzanine version 4.1.0, in 4.0.1 it worked, if i remember correctly.

molokov:
Duplicate of #1533, #1537 , fixed in #1516 
If you upgrade to master you should be fine. It's not in a PyPI release yet.

georgkrause:
@molokov Thank you! when does the new release come? is there a plan right now?

ryneeverett:
@georgkrause Releases are infrequent but master is generally quite stable.

georgkrause:
@ryneeverett i am not sure if "quite stable" is enough for a productive instance, especialy when using master just "should be fine". do you use the master in a productive instance?

ryneeverett:
@georgkrause I put master in production if there's a bugfix/feature I need, yes. It's a pretty slow moving project, so I can review every commit. I put my own code into production too and it doesn't even have the benefit of code review.

georgkrause:
got this problem again...

petedermott:
So I think I've fixed this. 

The issue is that its throwing an attribute error on models that are not searchable when trying to make the dictionary of urls for the admin links panel - quite why these are even getting used in the first place I'm still not sure.

My "solution" (hack) is just to catch and pass on these errors, these items shouldn't accessible for links anyway so it doesn't matter if our sitemaps just skip over them.

PR #1733, opened 2017-02-09
Try and fix admin link error

petedermott:
Fix TinyMCE link issue referenced #1728 and #1594

### DIFF
--- a/mezzanine/core/managers.py
+++ b/mezzanine/core/managers.py
@@ -399,10 +399,13 @@ class DisplayableManager(CurrentSiteManager, PublishedManager,
         setattr(home, "get_absolute_url", home_slug)
         items = {home.get_absolute_url(): home}
         for model in apps.get_models():
-            if issubclass(model, self.model):
-                for item in (model.objects.published(for_user=for_user)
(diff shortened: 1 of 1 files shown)


stephenmcd:
This was thought to be resolved and if so, you've presumably come across a similar yet different issue - there's no way of knowing though since you admittedly don't know what the cause is. Given that we can't just blindly accept a patch. Are you able to upload a minimal example with steps to reproduce the problem?

stephenmcd:
BTW, I'll close this since it most likely isn't the correct fix. Once the real issue can be identified, then we can work out what the correct fix is.

petedermott:
Understood. 

In hind sight more discussion and consistent replication of the issue probably would have been the better course of action, I just got over excited at finding a "fix".

I'm going to keep looking into this and will report my findings.

stephenmcd:
I understand, I've done this plenty of times when contributing to others' projects, so don't sweat it.

I'm glad you'll keep looking into it - I would suggest creating a new project from scratch, and adding your code to it piece by piece until you can trigger the error, that way you'll have a minimal example you can upload and someone else can take a look at it from there.

Issue #1400, opened 2015-09-04
Provide way to silence warnings


spookylukey:
We have a valid reason to not include  `"mezzanine.core.middleware.SitePermissionMiddleware"` in our `MIDDLEWARE_CLASSES` (we use an alternative implementation), but this produces a warning that we can't silence:

https://github.com/stephenmcd/mezzanine/blob/c7bc3e74d8487c6b23417ac099a46aa5c486c369/mezzanine/utils/sites.py#L75

It would be nice if Mezzanine used Django's checks framework for issues like this:

https://docs.djangoproject.com/en/1.7/topics/checks/

In this way, a developer could silence this warning individually.

I'm happy to write a patch if this idea is accepted. There are probably a few other places this could be done, although for things like deprecation warnings, there is no need - there is a right way to silence the warning, which is to fix your code.

stephenmcd:
Sounds good Luke, go for it.

jerivas:
Looks like #1734 fixes this one. Closing.

PR #1734, opened 2017-02-18
Use django check for SitePermissionsMiddleware

spookylukey:
See #1400 

I had a look, and I don't think that any of the other warnings that remain need this treatment — they are all things that need to be addressed in other ways, like deprecation warnings, rather than optional things that can be safely silenced.


stephenmcd:
Awesome, thankyou!

PR #1737, opened 2017-02-18
Remove request.REQUEST usage

spookylukey:
See https://docs.djangoproject.com/en/dev/releases/1.7/#merging-of-post-and-get-arguments-into-wsgirequest-request

I have manually checked that the login form works as expected with this change, I have not tested the search form as I didn't have a project set up that shows that, plus there aren't tests in the test suite at the kind of level necessary to show this kind of issue.


stephenmcd:
Thanks Luke!

PR #1738, opened 2017-02-23
Use requirements.txt as default REQUIREMENTS_PATH

andresmrm:
Since [local_settings.py says `requirements.txt` is the default value for `REQUIREMENTS_PATH`](https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/project_template/project_name/local_settings.py.template#L42), shouldn't it be set here?

Sorry if I should not be pushing this to `master`, but there are too many branches, so I didn't know which one to push.

### DIFF
--- a/mezzanine/project_template/fabfile.py
+++ b/mezzanine/project_template/fabfile.py
@@ -63,7 +63,7 @@ env.domains_python = ", ".join(["'%s'" % s for s in env.domains])
 env.ssl_disabled = "#" if len(env.domains) > 1 else ""
 env.vcs_tools = ["git", "hg"]
 env.deploy_tool = conf.get("DEPLOY_TOOL", "rsync")
-env.reqs_path = conf.get("REQUIREMENTS_PATH", None)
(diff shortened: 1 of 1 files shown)


stephenmcd:
Don't think this is right. There are checks throughout the fabfile that only try to install requirements if the value is defined. By giving it a default, those checks will always pass, even if the projct doesn't contains a requirements.txt file.

andresmrm:
At least for me here, I tried to run `fab create` and got an error because it didn't installed Mezzanine, because no `REQUIREMENTS_PATH` is set by default. Is that intended?

Shouldn't we change the comment at https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/project_template/project_name/local_settings.py.template#L42 to None then?
The way it is I understand `requirements.txt` is the default. Or maybe add more info to the comment?

Other possibility would be to add `mezzanine` here:
https://github.com/andresmrm/mezzanine/blob/master/mezzanine/project_template/fabfile.py#L535
But it doesn't seems right neither...

andresmrm:
Looks like I misunderstood all the comments there. They are not defaults. They need to be set for it to work. Sorry.

PR #1742, opened 2017-03-10
Defining version range for Bleach

### LINKED ISSUE #1741: Version 2.0 of Bleach is causing error
Seems like the new version of Bleach is incompatible with Mezzanine <=4.2.2.

After starting the dev server (`./manage.py runserver`) the following output can be seen:

```
    if "tel" not in sanitizer.BleachSanitizer.allowed_protocols:
AttributeError: module 'bleach.sanitizer' has no attribute 'BleachSanitizer'
```

Saving an object that requires sanitation is subsequently also causing an error:

```
ValueError at /de-de/admin/foo/bar/55/
attributes needs to be a callable, a list or a dict

Request Method:	POST
Request URL:	http://foobar.com/de-de/admin/foo/bar/55/
Django Version:	1.8.17
Exception Type:	ValueError
Exception Value:	
attributes needs to be a callable, a list or a dict
Exception Location:	/srv/http/foobar.com/envs/myenv_dev/local/lib/python2.7/site-packages/bleach/sanitizer.py in attribute_filter_factory, line 194
Python Executable:	/srv/http/foobar.com/envs/myenv_dev/bin/python
Python Version:	2.7.6
```

Downgrading to Version 1.5.0 fixes the issue. Perhaps Mezzanine should be pinned to Bleach 1.5.0 until the new version of Bleach can be used.

brejoc:
Bleach 2.0 is introducing breaking changes. We need something between version >= 1.5 and < 2.0.

Fixes #1741

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ try:
             "filebrowser_safe >= 0.4.6",
             "grappelli_safe >= 0.4.5",
             "tzlocal >= 1.0",
-            "bleach >= 1.4",
+            "bleach >= 1.4, <2.0",
             "beautifulsoup4 >= 4.1.3",
             "requests >= 2.1.0",
             "requests-oauthlib >= 0.4",


stephenmcd:
There was more involved, which I've resolved in 7009758b62694a3fbb7c48f7ee0aa978f293bd8a

Thanks all the same!

brejoc:
No worries. Thanks for fixing this!

PR #1750, opened 2017-04-12
Django 1.11 support

solarissmoke:
First pass at adding Django 1.11 support. Key changes:

- Always [pass a `dict`](https://docs.djangoproject.com/en/1.11/releases/1.11/#django-template-render-prohibits-non-dict-context) to `template.render`.
- Annotation of search results is now done explicitly. See [commit message](https://github.com/stephenmcd/mezzanine/commit/a7311e32636d0763f609b949ec420e85769c0af7) for explanation.
- Removed `null` attribute from `Slugged.slug`.

Comments:

- This could do with a bit of testing to make sure I haven't missed anything.
- There is an error in one of the tests (`test_password_reset`) which happens in master/1.10 as well. I can't figure out what is going on there.


stephenmcd:
This is a huge help, thanks a lot!

solarissmoke:
@stephenmcd I tracked down the cause of the `test_password_reset` error - it is an issue in `grappelli_safe`, which uses the `future` template tag in the `registration/password_reset_email.html` template. 

This has been fixed in the repo (the template was [removed ages ago](https://github.com/stephenmcd/grappelli-safe/commit/567f4aa0ccfc67fb000a3bf02ae077b996e17366) actually), but for some reason the package on pypi still has the template. Seems like that package might need to be fixed?

stephenmcd:
> Seems like that package might need to be fixed?

Thanks a lot for tracking that down - I'll try and push a new release of that soon.

melvyn-sopacua:
Feedback: currently developing with this PR and Django 1.11.1 and looking good.

stephenmcd:
Thanks @melvyn-sopacua

utapyngo:
Is it released yet?

PR #1752, opened 2017-04-21
Add explicit on_delete arguments to foreign keys in migrations.

solarissmoke:
In #1745 I forgot to apply the same changes to the existing migrations, which you also have to do.


stephenmcd:
Awesome, thanks!

PR #1754, opened 2017-04-24
Proposal for AppSettings object

melvyn-sopacua:
When accepted a new management command providing boilerplate can be used
to have settings picked up again automagically, by creating a
prepopulated ready() method that looks for `DefaultSettings` inside
defaults.py.

### DIFF
--- a/mezzanine/conf/__init__.py
+++ b/mezzanine/conf/__init__.py
@@ -114,8 +114,6 @@ class AppSettings(object):
 
     @classmethod
     def register(cls):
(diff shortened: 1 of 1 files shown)


stephenmcd:
I don't see what advantage this provides over the existing approach, but maybe it just needs further explaining.

We should also only have one approach, so the advantage to this would need to be vast in order to warrant replacing the existing approach. Given that, this can probably live outside of Mezzanine - so I'll close this for now. We can re-open if I'm completely off the mark.

PR #1755, opened 2017-04-26
Use TextField for Field.label and Field.help_text

### LINKED ISSUE #1753: Form field Help Text is too short
Right now the form fields created with `mezzanine.forms` limit the `help_text` to just 100 characters (see [forms/models.py](https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/forms/models.py#L68)). Form fields that require a disclaimer or lengthy explanation are being cut off in the help text, so my users have been working around it by putting all the text in the label (which has a limit of 200 and can be raised above that).

What do you think of making the `help_text` limit configurable?

jerivas:
Fixes #1753.

The main things are:
- `settings.FORMS_LABEL_MAX_LENGTH` will no longer trigger a migration when changed.
- A new setting `FORMS_HELP_TEXT_MAX_LENGTH` was created to allow for configurable `help_text` max length. It will not trigger migrations when changed by the user.

It turns out that implementing the max_length checks was a bit trickier than expected (because it requires modifying the field, the widget, and adding a validator). Also, we're modifying the fields in `__init__` instead of redefining them completely to inherit as much as possible from the model (verbose name, blank flag, etc).


jerivas:
Are new settings added to the docs by some automated build step? Or should I add it manually?

stephenmcd:
I feel like we should get rid of the previous and new setting. The original setting only existed due to CharFields requiring a value, that would always eventually be too short for someone. I don't think having those settings merely as validators for the fields is warranted. If the user really wants to set restrictions for those they can probably path them on somehow, but I don't think that will be a common requirement.

stephenmcd:
> Are new settings added to the docs by some automated build step?

Yep that happens here: https://github.com/stephenmcd/mezzanine/blob/master/docs/conf.py#L39

jerivas:
You're right, and removing the settings simplifies this a lot.

Regarding the docs, does that mean there's a command to be run for them to rebuild? Or is it something you do upon pushing a new release? Sorry, I couldn't find anything in the contributing section.

stephenmcd:
> does that mean there's a command to be run for them to rebuild?

Yeah it's just the sphinx command for building them, nothing special required, it will pick up everything in docs/conf.py:

`sphinx-build docs docs/build`

stephenmcd:
BTW thanks for working on this, much less painful now :-)

jerivas:
Always happy to help :)

For future reference of anyone finding this:

- `settings.FORMS_LABEL_MAX_LENGTH` has been removed. Field labels don't have a length limit now.
- Field help_texts also don't have a length limit.

PR #1763, opened 2017-05-18
Weigh search results by age

simenheg:
Weigh search results by their age by default. Add a new setting, `SEARCH_AGE_SCALE_FACTOR`, controlling how much emphasis to put on the age when ranking results (set this to 0 to revert to the old behavior).

### DIFF
--- a/docs/search-engine.rst
+++ b/docs/search-engine.rst
@@ -128,14 +128,15 @@ Ordering Results
 ================
 
 By default, results are ordered by the number of matches found within the
-fields searched. It is possible to control the relative weight of a match
-found within one field over a match found in another field. Given the
-first example of searching :class:`.Page` instances, you might decide that
-a match within the ``title`` field is worth 5 times as much as a match in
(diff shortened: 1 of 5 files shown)


stephenmcd:
Thanks a lot for this, I think it's a great idea! I think the implementation could be better though:

- No need for a setting to disable it, just have it always on 
- No need for the "6 month" drop off, have a look at the time scaling algorithm here: https://github.com/stephenmcd/drum/blob/c92eff9952b340256e1c02ee92151630ecbf2986/drum/links/utils.py#L48-L50

simenheg:
Thanks your feedback, @stephenmcd. I've updated the commit addressing your comments.

I did adjust the default the default scaling factor from 2 to 1.5 because a factor of 2 seemed to bring too many irrelevant results to the first page because of their freshness in my testing, while 1.5 seemed to provide a good balance.

stephenmcd:
Really awesome, thanks again!

PR #1766, opened 2017-05-31
Modified the blog homepage pattern to a working version. Added import it requires.

vranki:
Currently showing blog as front page requires knowledge of Mezzanine internals as a new import is required to the urls.py. This is way too difficult for a non-developer.

This commit adds the required import and modifies the related url pattern. Now user only needs to uncomment the url pattern line (and comment previous) and the import line to get the blog visible.

Not tested properly with git version and I'm not familiar with the codebase so please test before merging. Works on 4.2.3. ./setup.py test fails due to unrelated errors.


stephenmcd:
Thanks

PR #1767, opened 2017-06-02
Remove explicit show_banner argument from inner_run

Ivoirians:
This broke --noreload for me. When called from https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L107, if **options contains a show_banner arg (which, seemingly, is set just above in add_arguments), this throws a "Got multiple values for argument" TypeError.

Tried (and failed) to figure out why I might have this issue and no one else seems to--but this should be a pretty innocuous change.


stephenmcd:
Thanks!

Issue #1457, opened 2015-11-05
Multible languages in the admin settings tabs are buggy


henri-hulski:
The translatable fields in the settings tab (`Tagline` and `Site title`) show the values for all languages in the language the admin uses and not the values from the database.

For example I save `Site Title` in English as `Shop` and in French as `Boutique`.
This is correctly stored in the database.
When I reload the admin if using `/en/admin/`
it shows for French and English `Shop`, but if using `/fr/admin/` it shows for French and English `Boutique`.

When I change now something else in the `Settings` tab, when saving it stores the values for `Tagline` and `Site title` which are currently shown in the database.

henri-hulski:
Another related issue is, that `Order Email Subject` field should be translatable.

henri-hulski:
For me this is really an annoying bug. 
Always when I change something in the Admin settings, I have to restore the translations of `Tag line`,  `Site title` and `Order Email Subject` by hand with `pgAdmin3`. 

@Kniyl
Do you have an idea where and how to fix this bug? 
I could try to fix it but I have no idea where to look.

KagelVlusha:
Hey, I've run into this too, and I'm pretty sure the problem is in mezzanine.conf.forms.SettingsForm._init_field. On line 60, the initial value for these forms is set to gettatr(settings, name) which, because of the magic in django_modeltranslations, always returns the value for the current language.

Also, there's another minor bug where any non-translated settings appearing after a translated setting on this page will be displayed in the last language in settings.LANGUAGES. This is because on line 50 in __init__ for the same form object, activate(active_language) is outside the preceding for loop, and so the language only gets reset to the current language after all fields have been initialized, rather than after each field.

henri-hulski:
@KagelVlusha Do you want to create a Pull Request? I'm not working with Mezzanine or Django atm.

KagelVlusha:
Pull request created. It's #1773.

stephenmcd:
Thanks for this.

At a glance, this looks like it mightn't be optimal as it appears to run a DB query for every setting. Is there some solution that clears the settings cache and forces them to be loaded as they normally are from the DB? (which I believe is done much more efficiently with a single query.)

PR #1773, opened 2017-06-15
Fix for issue #1457

KagelVlusha:
Fixed two problems (from issue #1457) with the editable settings admin when settings.USE_MODELTRANSLATION is True.

1) The settings for languages other than the user's current language used the wrong initial values
2) Non-translated settings appearing bellow a translated setting were displayed in the wrong language

### DIFF
--- a/mezzanine/conf/forms.py
+++ b/mezzanine/conf/forms.py
@@ -47,17 +47,23 @@ class SettingsForm(forms.Form):
                             self._init_field(setting, field_class, name, code)
                 else:
                     self._init_field(setting, field_class, name)
-        activate(active_language)
+            activate(active_language)
 
     def _init_field(self, setting, field_class, name, code=None):
         """
         Initialize a field whether it is built with a custom name for a
         specific translation language or not.
         """
+
+        if settings.USE_MODELTRANSLATION and setting["translatable"]:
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks for this.

At a glance, this looks like it mightn't be optimal as it appears to run a DB query for every setting. Is there some solution that clears the settings cache and forces them to be loaded as they normally are from the DB? (which I believe is done much more efficiently with a single query.)

KagelVlusha:
The problem with clearing and reloading the cache is that the way the settings admin is currently set up, any translatable setting requires text in all available languages be loaded, and the cache is only set up to hold one language, the user's current language, at any given time. 

If we want to be query efficient about this, then the system behind how settings are cached would have to be reworked so that the cached object for translatable settings is more than just a string, but an object or a dictionary or something with all possible languages.

I could take a crack at writing something like that, but personally I don't think it's needed. There is no other use case for it that I'm aware of, and the inefficiency here really only amounts to ~10 extra queries on one admin page. Sure it's not optimal, but unless you've set up a project with hundreds of custom translatable settings, I don't think it'll make the page unusable.

stephenmcd:
Makes sense. Thanks again.

PR #1774, opened 2017-06-28
Correction to pull request #1773 (Fix for issue #1457)

KagelVlusha:
My apologies, but I've just discovered a bug in the code I submitted in pull request #1773.

It occurs if there are no Setting objects in your database, so typically only the first time the settings admin is used. If you modify an existing project with the previous version of this code, like I did when testing it, it works fine. However, when creating a new project, the settings admin will throw an error and refuse to load at all.

Again, my apologies.

### DIFF
--- a/mezzanine/conf/forms.py
+++ b/mezzanine/conf/forms.py
@@ -44,21 +44,26 @@ class SettingsForm(forms.Form):
                         except:
                             pass
                         else:
-                            self._init_field(setting, field_class, name, code)
+                            self._init_field(setting, field_class, name, code,
+                                             active_language,)
                 else:
                     self._init_field(setting, field_class, name)
             activate(active_language)
 
-    def _init_field(self, setting, field_class, name, code=None):
+    def _init_field(self, setting, field_class, name, code=None,
+                    active_language=None,):
         """
         Initialize a field whether it is built with a custom name for a
         specific translation language or not.
         """
 
-        if settings.USE_MODELTRANSLATION and setting["translatable"]:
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks. 

These changes cause some new test failures since we run syntax test to ensure code is less than 80 chars per line. Can you wrap those so that those tests pass?

KagelVlusha:
OK, should be done now.

stephenmcd:
Thanks a lot - sorry to nitpick though, but I think you've changed the formatting to extremely by placing every element onto its own line. We generally just wrap lines only as needed, for example:

Here's a long if statement wrapped, you can see as many elements occur per line as possible, and it wraps only when necessary:
https://github.com/stephenmcd/mezzanine/blob/920bdae9f5745ef79ff08731ad603a18ffe9edeb/mezzanine/core/middleware.py#L199-L200

Same for function/method definitions:
https://github.com/stephenmcd/mezzanine/blob/920bdae9f5745ef79ff08731ad603a18ffe9edeb/mezzanine/utils/views.py#L162-L163

Lastly, since all these extra changes have occurred coupled with relatively non-descriptive commit messages that will form part of the change log for the next release, could you squash these changes into a single commit with a descriptive commit message? Something like "Fixed handling of translatable settings in admin when no settings in DB" is a lot more useful than "Fixed a bug".

Thanks again.

KagelVlusha:
And done.

stephenmcd:
Thank you very much!

PR #1783, opened 2017-07-28
Remove device detection and `mezzanine.mobile`

jerivas:
This removes all code, tests, and docs related to device detection and the `mezzanine.mobile` package, which contained mobile-specific templates.

Users upgrading from previous versions should see a warning if they have `TemplateForDeviceMiddleware` in their `MIDDLEWARE` settings. They can simply remove it to silence it. A warning will also be shown if `mezzanine.mobile` is included in `INSTALLED_APPS`.

I kept the doc page for device detection with a note saying the feature was removed in Mezzanine 4.3 (assuming that's the release where this is going to land).

Would like to hear thoughts from the community.


mezzanine/utils/cache.py
```diff
@@ -86,12 +85,11 @@ def flatten(seqs):
...
     """
     Cache key for Mezzanine's cache middleware. Adds the current
-    device and site ID.
+    site ID.
     """
-    cache_key = "%s.%s.%s." % (
+    cache_key = "%s.%s." % (
         settings.CACHE_MIDDLEWARE_KEY_PREFIX,
         current_site_id(),
-        device_from_request(request) or "default",
```
stephenmcd: Maybe we just leave the string "default" here with a note in the comment - I think for most people that would mean their cache keys stay the same.

docs/device-handling.rst
```diff
@@ -2,74 +2,5 @@
...
-
-==================================================  ==================================================
-Django                                              Mezzanine
-==================================================  ==================================================
-``django.shortcuts.render``                         :func:`mezzanine.utils.views.render`
-``django.template.Library().inclusion_tag``         :meth:`mezzanine.template.Library().inclusion_tag`
-``django.views.generic.simple.direct_to_template``  :func:`mezzanine.core.views.direct_to_template`
-==================================================  ==================================================
+Mezzanine used to include the ability to load different templates according to
+the request's user agent. This feature was removed in Mezzanine 4.3.
```
stephenmcd: How about linking to an alternative, like https://github.com/gregmuellegger/django-mobile

docs/index.rst
```diff
@@ -19,12 +19,10 @@ powered by Mezzanine <http://mezzanine.jupo.org/sites/>`_.
     Mezzanine Is Just Django* - `Ken Bolton <http://bscientific.org/>`_,
     Mezzanine core team member.
 
-**Front-end developers** can read about how to set up templates for
-specific :doc:`device-handling` such as phones and tablets. Mezzanine
-also comes with the ability for content authors to edit content directly
-within a page while viewing it on the website. You can read about this
-and how to implement this feature within templates under
-:doc:`inline-editing`.
+**Front-end developers** migh be interested in Mezzanine's ability for
```
stephenmcd: Might*

jerivas:
Just pushed some changes addressing your comments.

stephenmcd:
Awesome, thanks again!

PR #1784, opened 2017-07-28
Handle host-themes via a template loader

jerivas:
This implements host-specific templates using a template loader instead of a middleware. Host-detection now works on all templates, like those included by templatetags or by `{% include %}` or `{% extend %}` calls. The middleware only worked on the final template rendered by a view.

I've also included a terminal warning with instructions for old projects to upgrade to the loader approach.

### DIFF
--- a/docs/multi-tenancy.rst
+++ b/docs/multi-tenancy.rst
@@ -133,3 +133,38 @@ an independent page hierarchy, etc.). Furthermore, the content types
 added to, say ``example_theme``, e.g. ``HomePage``, are shared and
 available in the different sites. Such nuances of sharing must be
 considered when employing this approach.
+
+Upgrading from ``TemplateForHostMiddleware``
+--------------------------------------------
+
+Mezzanine implements host-specific templates using a template loader since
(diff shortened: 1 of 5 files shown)


stephenmcd:
Should this also include removing/deprecating `mezzanine.template.Library.inclusion_tag` ?

stephenmcd:
Also wondering if that `Library` class is entirely redundant yet?

jerivas:
The `inclusion_tag` method was removed in #1783. I can explore getting rid of `Library` altogether. My understanding was that it's needed for the `{% end* %}` templatetags.

stephenmcd:
> The inclusion_tag method was removed in #1783. 

Sorry, missed that.

> I can explore getting rid of Library altogether. My understanding was that it's needed for the {% end* %} templatetags.

I kinda thought the same, just wanted to make sure.

stephenmcd:
Thanks Ed, it's such a better approach.

jerivas:
Always happy to help 😄

PR #1792, opened 2017-08-23
Fix collecttemplates

melvyn-sopacua:
Properly parse positional args

As demonstrated by using call_command, the positional args are
not processed. This may be a left-over from optarg migration.

Usage property is no longer necessary (and ignored even).

Without patch (python 3.6 virtualenv):

    usage: manage.py collecttemplates [-h] [--version] [-v {0,1,2,3}]
                                  [--settings SETTINGS]
                                  [--pythonpath PYTHONPATH] [--traceback]
                                  [--no-color] [--noinput] [-t TEMPLATE] [-a]
    manage.py collecttemplates: error: unrecognized arguments: nonexistent


melvyn-sopacua:
Needs one more fix:

The comparison with INSTALLED_APPS is a shortcut, but with the new apps framework will not work for apps that follow the new (and recommended) approach to reference an AppConfig derived class.

Also some thoughts for enhancement in #1793.

stephenmcd:
Thanks

PR #1798, opened 2017-09-05
Add public visibility status on Displayable model

simenheg:
Add a method to the Displayable model for determining whether the instance is currently visible to the public. This can for instance be useful when wanting to style non-public pages or posts differently to make admin users able to tell them apart form public ones.

### DIFF
--- a/mezzanine/core/models.py
+++ b/mezzanine/core/models.py
@@ -274,6 +274,15 @@ class Displayable(Slugged, MetaData, TimeStamped):
         return timesince(self.publish_date)
     publish_date_since.short_description = _("Published from")
 
+    def is_public(self):
+        """
+        Return ``True`` if this visible to the public, or ``False`` if it's
+        only shown for admin users.
+        """
+        return self.status == CONTENT_STATUS_PUBLISHED and \
+            (self.publish_date is None or self.publish_date <= now()) and \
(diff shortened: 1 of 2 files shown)


molokov:
I'd like to +1 this change... I think it's quite useful for an admin to know whether or not a page is viewable to the public (more than once, I've publicised a page without remember it was a) in draft or b) the expiry date was in the past).

I had added this to my base.html template to show "DRAFT" or the expiry date at the top left of the page, but with an "is_public()" method, the code for this can easily be simplified (although admittedly the information shown would simply be "NOT PUBLIC" or something)

`{% if user.is_staff or user.is_superuser %}
{% if page.get_content_model.status == 1 or page.get_content_model.expiry_date %}
<p class=" text-left label label-success" style="position:fixed; left:0; top:80;">
{% if page.get_content_model.status == 1 %}DRAFT<br/>{% endif %}
{% if page.get_content_model.expiry_date %}Expires {{page.get_content_model.expiry_date | date:"d-m-y H:i"}}<br/>{%endif%}
</p>
{% endif %}{% endif %}
`

stephenmcd:
Thanks for this. I think it should be `published` instead of `is_public` since we already use that name to represent that same thing on the manager. I'll merge and make that change myself.

PR #1801, opened 2017-09-09
Added more tests por pages

vindio:
Added more tests for mezzanine pages to increase code coverage.
Found a bug in models_for_pages (commented test fails with "PageAdmin.get_content_models() requires self argument")


stephenmcd:
Thanks for this! Could you either remove the commented code or fix the issue with it?

vindio:
Not sure about how to fix the bug so I removed the commented test

vindio:
Not sure about how to fix the bug so I removed the commented test

stephenmcd:
Thank you!

PR #1803, opened 2017-09-18
JPEG Handling changes in PIL/Pillow

zandeez:
Resolves the following:

* JPEG images cannot be saved as RGBA
* JPEG images may still be of a mode that we want to convert from
* Padded images are always created RGBA which breaks JPEG

### DIFF
--- a/mezzanine/core/templatetags/mezzanine_tags.py
+++ b/mezzanine/core/templatetags/mezzanine_tags.py
@@ -387,10 +387,12 @@ def thumbnail(image_url, width, height, upscale=True, quality=95, left=.5,
         to_width = from_width * to_height // from_height
     elif to_height == 0:
         to_height = from_height * to_width // from_width
-    if image.mode not in ("P", "L", "RGBA") \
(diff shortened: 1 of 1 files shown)


stephenmcd:
I'm able to successfully generate thumbnails for jpegs without this change, can you provide more info on how to reproduce the problem?

stephenmcd:
I still can't reproduce the problem. I used the vars you specified for the `thumbnail` tag and I get a valid resized image, see: 

screen shot 2018-05-15 at 11 42 04 am

viaregio:
If I remember correctly, the problem zandeez is describing occurs with newer versions of PIL/pillow.

zandeez:
Seems to be resolved Mezzanine 4.3.1, Pillow 5.3.0

zandeez:
Ignore me, the patch does work I hadn't applied it correctly.

Basically with the versions listed above, if thumbnailing with padding with a format that doesn't support RGBA (such as jpg) the thumbnail will not save. This patch forces jpegs to be RGB rather than RGBA which saves consistently.

Issue #1805, opened 2017-09-23
500 Error in Admin (admin/login) when CachedFilesMixin is used


awartani:
```
class NonPackagingS3PipelineCachedStorage(
        NonPackagingMixin, PipelineMixin, CachedFilesMixin, S3BotoStorage):
    location = 'static'
```

```
STATICFILES_STORAGE = 'storage.NonPackagingS3PipelineCachedStorage'
```

with `DEBUG=True`

Will lead to a 500 error. 

Debugging it for a while, it turned out that the root cause of this
```
window.__admin_media_prefix__ = '{% static "admin" %}/';
```
https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/templates/admin/base_site.html#L21

because it tries to use Static engine to convert a directory to a file.

stephenmcd:
As per #1806

PR #1806, opened 2017-09-23
Use STATIC_URL to find window.__admin_media_prefix__

awartani:
Related to https://github.com/stephenmcd/mezzanine/issues/1805

### DIFF
--- a/mezzanine/core/templates/admin/base_site.html
+++ b/mezzanine/core/templates/admin/base_site.html
@@ -18,7 +18,7 @@
     window.__tinymce_css = '{% static "mezzanine/css/tinymce.css" %}';
     window.__admin_url = '{{ admin_index_url }}';
     window.__static_proxy = '{{ static_proxy_url }}';
-    window.__admin_media_prefix__ = '{% static "admin" %}/';
+    window.__admin_media_prefix__ = '{{ STATIC_URL }}';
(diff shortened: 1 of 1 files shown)


stephenmcd:
Presumably this breaks since `STATIC_URL` doesn't end with "/admin/"

awartani:
@stephenmcd good catch, what if it was changed to `window.__admin_media_prefix__ = '{{ STATIC_URL }}/admin/';`

stephenmcd:
I think the problem's with whatever staticfiles class you're using.

awartani:
@stephenmcd I don't think so, the details are on https://github.com/stephenmcd/mezzanine/issues/1805 if you would like to see what's the issue.

Anyway, `'{% static "admin" %}/'`static is used to look up for a directory name, not a file and that where it crashed when it was linked to S3.

stephenmcd:
What I mean is it works fine with the default staticfiles class, which indicates a problem with the one you're using.

awartani:
@stephenmcd I don't think it's wise to think about it that way and to close the PR.

ryneeverett:
@awartani If you demonstrated a clear understanding of the bug you're trying to fix you would probably be happier with the response. Stephen went with his intuitions because your technical arguments have been vague.

PR #1807, opened 2017-09-27
Test Nginx config before restarting

andresmrm:
This way the configuration won't be updated if it's broken, and you also get an error message in the terminal explaining why it's broken (instead of just telling you to check service status/journal).


stephenmcd:
Thanks!

PR #1810, opened 2017-10-02
Enable browser-native spell checking in WYSIWYG tinymce editor, which…

mwort:
… got wiped in 82339b0 . Previously introduced in 86f6ef6 .

This seems to have been disabled when upgrading support for Tinymce 4. I used the more general `browser_spellcheck` rather than the `gecko_spellcheck` attribute used previously. Tested in Safari and Firefox.


stephenmcd:
Thanks!

PR #1812, opened 2017-10-06
Fixed the bug with the response header and non-ascii cache names

Kianoosh76:
Hi, I faced a weird problem when I tried to comment with a non-ascii name and made a solution for it with a simple trick.

The story is explained [here](https://stackoverflow.com/questions/46605315/django-mezzanine-comments-with-non-ascii-cache-names) in much more details. The solution comes from the first comment.

### DIFF
--- a/mezzanine/generic/views.py
+++ b/mezzanine/generic/views.py
@@ -110,7 +110,7 @@ def comment(request, template="generic/comments.html", extra_context=None):
         # Store commenter's details in a cookie for 90 days.
         for field in ThreadedCommentForm.cookie_fields:
             cookie_name = ThreadedCommentForm.cookie_prefix + field
-            cookie_value = post_data.get(field, "")
+            cookie_value = post_data.get(field, "").encode('ascii', errors='ignore')
             set_cookie(response, cookie_name, cookie_value)
         return response
     elif request.is_ajax() and form.errors:


stephenmcd:
Can you add a test that fails without this change, and passes with it?

stephenmcd:
And thanks a lot!

Kianoosh76:
I tried to add a test but actually the error is raised by the web server! Both nginx and django local server raise the error in their own manner.
So the response is returned apparently with no errors (and the test passes) but it contains non-ascii characters in its header which causes the web server to fail encoding it.

stephenmcd:
Part of the need for a test is it will allow the reviewer to confirm the bug. It sounds like you might in fact have an issue with how the locale is configured for your environment. This has been discussed ad nauseam on the mailing list before, can you see if that's your issue? Please re-open if you can validate that the code actually does need changing.

Kianoosh76:
I actually don't know how to access the mailing list but confirming this bug is pretty straight-forward. Just put 'س' in the comment author name and you'll (probably?) see that the server responses with error 500 when trying to encode response header.

This is also explained [here](https://tools.ietf.org/html/rfc7230#section-3.2.4):
   ```
Historically, HTTP has allowed field content with text in the
   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
   through use of [RFC2047] encoding.  In practice, most HTTP header
   field values use only a subset of the US-ASCII charset [USASCII].
   Newly defined header fields SHOULD limit their field values to
   US-ASCII octets.  A recipient SHOULD treat other octets in field
   content (obs-text) as opaque data.
```
This rule may be violated by putting the author name in the response header (for caching) because it may contains non-ascii characters

By the way, if the problem will be solved by changing locale settings, I appreciate any other solutions!

stephenmcd:
Works for me:

screen shot 2017-11-04 at 4 38 21 pm

Kianoosh76:
That's weird! Was you running the project on the django local server? Now it seems the problem is not by your side! BTW as I mentioned above, the standard behavior is not to put non-ascii characters in the response header.

stephenmcd:
Works for me using Django's development server, and also production gunicorn/nginx.

PR #1815, opened 2017-10-19
Don’t weigh search results in the future

simenheg:
Fix a bug occurring when search results have publish dates in the future. Before this patch, an unlucky combination of a negative age and a fractional `SEARCH_AGE_SCALE_FACTOR` would result in a complex number search weight, giving problems with sorting since floats and complex numbers cannot be compared.

Simply ignore the age of future results during weighing instead.

### DIFF
--- a/mezzanine/core/managers.py
+++ b/mezzanine/core/managers.py
@@ -219,7 +219,8 @@ class SearchableQuerySet(QuerySet):
 
                 if result.publish_date:
                     age = (now() - result.publish_date).total_seconds()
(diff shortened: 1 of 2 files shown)


stephenmcd:
Should the fix involve making the types consistent and comparable?

simenheg:
I'm not sure that I understand your question.

Complex numbers can't be compared meaningfully, so assigning complex number weights was a bug introduced by an oversight on my part. The weights should always be real numbers, which is what this patch tries to ensure.

stephenmcd:
Sorry I misunderstood the original issue, thanks for the fix.

Issue #1814, opened 2017-10-12
In-line Editing: If editable text is white, it comes through the in-line editing form white as well, making the form appear empty


awright91:
Hello!
I've only been playing around with Mezzanine for a few weeks(loving it so far), but have been working with Django for a while now. I really enjoy Mezzanine's in-line editing, but I found one issue. 

Scenario: Let's say you're building a landing page with a full-width image and you want the text overtop of the image to be editable. The image is pretty dark, so you want the text to be white, like in the image below:

![image](https://user-images.githubusercontent.com/20651992/31519880-f0971da0-af71-11e7-90cc-e89375b9b377.png)

When you click the Edit button, the form appears empty, but only because the text is white, as you can see in this image: 

![image](https://user-images.githubusercontent.com/20651992/31519980-48684414-af72-11e7-9915-30460b6185e9.png)

I found a very simple one-line CSS fix, so if you all don't mind, I'd love to submit a PR for this. 

I've been wanting to contribute to some open source projects and I think Mezzanine is really awesome, so I'd love for this to be my first open source contribution, if it's cool with you all. 

Let me know and I'll submit a PR!

stephenmcd:
Please go for it. Note there's a CSS file specifically for this here:

https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/static/mezzanine/css/editable.css

PR #1816, opened 2017-10-19
Update editable.css

awright91:
Updated to fix the color issue of the in-line editing form. If color of text being edited is white, text comes through the form white, making it appear empty.

Issue explained with images here: https://github.com/stephenmcd/mezzanine/issues/1814

You can also test this issue on the demo site using the chrome dev tools.

I simply added an explicit color (#000) to the form text, ensuring consistency no matter what color the text being edited is.


stephenmcd:
Thanks

PR #1836, opened 2018-01-02
Update deprecated import references for django 2 using compat.py pattern

klowe0100:
Hi. Thanks for your great project! 

I recently had to unblock myself upgrading to django 2.0.  I'm up for spending some dev time to polish the changes more if you would like, including some regression testing.

### DIFF
--- a/mezzanine/compat.py
+++ b/mezzanine/compat.py
@@ -1,0 +1,32 @@
+import django
+
+
+_django_ver = django.VERSION
+
+is_dj2 = (_django_ver[0] == 2)
+
+if is_dj2:
+    from django.forms.widgets import SelectDateWidget
+    from django.urls import (resolve, reverse, NoReverseMatch,
+                             get_script_prefix)
+
+else:
+    from django.forms.extras.widgets import SelectDateWidget  #noqa
+    from django.core.urlresolvers import (resolve, reverse, NoReverseMatch, #noqa
+                                          get_script_prefix)
+
+
+def get_rel_to(self):
+    if is_dj2:
+        return self.remote_field
+    else:
+        return self.rel.to
+
+
+def get_js_catalog():
+    if is_dj2:
+        from django.views.i18n import JavaScriptCatalog
+        return JavaScriptCatalog.as_view()
+    else:
+        from django.views.i18n import javascript_catalog
+        return javascript_catalog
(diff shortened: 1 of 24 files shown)


jerivas:
Thanks for this! Would you also be interested in providing pull requests for filebrowser-safe and grappelli-safe (the two packages Mezzanine depends on)?

mezzanine/core/managers.py
```diff
@@ -175,8 +175,9 @@ def _clone(self, *args, **kwargs):
         """
         Ensure attributes are copied to subsequent queries.
         """
-        for attr in ("_search_terms", "_search_fields", "_search_ordered"):
-            kwargs[attr] = getattr(self, attr)
+        # i'm not quite sure what this is intended to do.
+        # for attr in ("_search_terms", "_search_fields", "_search_ordered"):
+            #kwargs[attr] = getattr(self, attr)
         return super(SearchableQuerySet, self)._clone(*args, **kwargs)
```
stephenmcd: I'd err on the side of this code actually doing something, let's not just blindly remove it.

Also please don't leave commented out code (mostly found with imports in this change).

mezzanine/compat.py
```diff
@@ -0,0 +1,32 @@
+import django
+
+
+_django_ver = django.VERSION
+
+is_dj2 = (_django_ver[0] == 2)
+
+if is_dj2:
```
vstoykov: Normally in Python is [easier to ask for forgiveness then ask for permission](https://docs.python.org/3.6/glossary.html#term-eafp). For that reason here is better to use `try:..except:` instead of checking if this is Django 2 or not.

mezzanine/core/admin.py
```diff
@@ -8,7 +8,8 @@
 from django.contrib.auth.models import User as AuthUser
 from django.contrib.redirects.admin import RedirectAdmin
 from django.contrib.messages import error
-from django.core.urlresolvers import NoReverseMatch
+from mezzanine.compat import NoReverseMatch
+# from django.core.urlresolvers import NoReverseMatch
```
vstoykov: As @stephenmcd said already you should not leave commented code. In your case you should just replace the line with it's new version. If you want to see the old version of that line there is GIT. It's exactly for that reason.

vstoykov:
I think that this should be closed now because in master there are already fixes for Django 2+ which also drop support for Django < 1.11 and the compatibility shims introduced here are not needed any more.

PR #1839, opened 2018-01-05
Implement optional group permissions for pages

mwort:
Mezzanine currently does not support per-page group viewing permissions, leaving the Group model only effective for staff users. Two mezzanine add-on packages addressed this ([mezzanine-protected-pages](https://github.com/simodalla/mezzanine_page_auth) and [mezzanine-page-auth](https://github.com/simodalla/mezzanine_page_auth)). However, both are no longer updated and incompatible with the current master, presumably because the overhead required didn't justify the small added functionality. Both needed an extra app entry in `INSTALLED_APPS`, relied on extra models and separate admins which seems over-complex for what it achieves in the end.

I'm proposing an optional `permitted_groups` field to the Page model similar to and in addition to the `login_required` to allow more granular page permissions. Since `login_required` is already handling most of this, only few lines are needed. The `PAGE_GROUP_PERMISSIONS` setting enables or disables this without hick-ups when it is set to `False` again.

A test is included. I'm happy to add some more documentation although the settings description already describes this quite well.


mwort:
I forgot to mention that this relies on `QuerySet.intersect` which was added in Django 1.11. Could probably be backported with `set()`.
I cant reproduce the error in `test_editable_caching` that Travis is reporting under Python 2.7 tho.

mwort:
Tests are passing for all python version except 2.7 now, I cant reproduce the error in `test_editable_caching` (`del _thread_local.request: AttributeError: request`) under OSX 10.13 or SLES 12.1, but it has occurred in previous PRs.

stephenmcd:
My preference would be for work to continue in a separate library - if those you mentioned aren't being maintained, then you could extend one of those and we can update links in the documentation to it.

stephenmcd:
Closing for now. We can re-open if strong opposition.

PR #1852, opened 2018-04-03
Fix sense of SITE_PERMISSION_MIDDLEWARE check

crazyscot:
This rectifies an issue introduced by commit 00f4a63c which incorrectly inverted the sense of the check for SITE_PERMISSION_MIDDLEWARE, causing a spurious mezzanine.core.W04 warning.

### DIFF
--- a/mezzanine/core/checks.py
+++ b/mezzanine/core/checks.py
@@ -160,7 +160,7 @@ def _build_suggested_template_config(settings):
 @register()
 def check_sites_middleware(app_configs, **kwargs):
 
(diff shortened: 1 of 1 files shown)


stephenmcd:
Thanks a lot

spookylukey:
In addition, it looks like my commit also inverted the earlier check for SITE_PERMISSION_MIDDLEWARE - https://github.com/stephenmcd/mezzanine/commit/00f4a63cb10cd27120571c6c84a13f3617ecea78#diff-9ca8dc28626fda77028150b232523679R94 - which has worse consequences of changing the behaviour. My apologies!

crazyscot:
Further PR on its way...

PR #1853, opened 2018-04-04
Fix sense of other check for SITE_PERMISSION_MIDDLEWARE.

crazyscot:
Issue introduced by commit 00f4a63. This follows from the thread on #1852.


stephenmcd:
Thanks again

PR #1854, opened 2018-04-09
Fix TemplateSettings unicode/str

spookylukey:
The main motivation behind this is to stop the warning:

    UserWarning: __unicode__ is not in TEMPLATE_ACCESSIBLE_SETTINGS

that appears if you are using django-debug-toolbar. This is caused ultimately by defining `__getattr__` which ends up being called via `hasattr(obj, "__unicode__")`.

In addition, this PR adds some basic tests for TemplateSettings, and makes the repr much more useful/accurate than before (which helps with debugging e.g. using django-debug-toolbar templates panel).

### DIFF
--- a/mezzanine/conf/context_processors.py
+++ b/mezzanine/conf/context_processors.py
@@ -2,11 +2,14 @@ from __future__ import unicode_literals
 
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


stephenmcd:
Thanks Luke

PR #1855, opened 2018-04-12
Support access to related model on Django 2+

wm3ndez:
`field.rel` was removed in Django 2.  These changes maintain compatibility with Django<2


stephenmcd:
Awesome, thank you.

PR #1856, opened 2018-04-20
Make thumbnail tag recognize .PNG and .GIF

simenheg:
Files with the upper case extensions .PNG and .GIF are now recognized by the thumbnail template tag as being PNG- and GIF images, respectively, instead of being treated as JPEG images.


stephenmcd:
Thanks

PR #1858, opened 2018-05-16
Support `SelectDateWidget` in django 1.8 and django 2.x

wm3ndez:


mezzanine/core/forms.py
```diff
@@ -5,7 +5,12 @@
 from uuid import uuid4
 
 from django import forms
-from django.forms.extras.widgets import SelectDateWidget
+try:
+    from django.forms.widgets import SelectDateWidget
+except:
```
kenbolton: Perhaps nit-picking, but I don't like bare `except`. How about `except ImportError:` across all three files?

stephenmcd: +1 please add ImportError

wm3ndez:
@kenbolton @stephenmcd Done. Thanks.

Issue #1313, opened 2015-05-25
Problem with ADMIN_REMOVAL and mezzanine.boot.lazy_admin.lazy_registration


jammon:
I have problems with making `ADMIN_REMOVAL` work. It looks like this results from `mezzanine.boot.lazy_admin.lazy_registration` being called before the code in `mezzanine/urls.py` that processes `ADMIN_REMOVAL`.
I started this as a [question on stackoverflow](http://stackoverflow.com/questions/30433029/how-to-use-admin-removal-in-mezzanine), which I copy here:

In my new mezzanine site I want to remove some models from the admin menu. So I defined `ADMIN_MENU_ORDER` with the models I want to see in the admin menu. This results in the other models, that are not in the list (`BlogPost`, `ThreadedComment`, `Site`, `Redirect`), being listed below my desired models.

So I used this to remove them:

    ADMIN_REMOVAL = ('blog.BlogPost',
                     'generic.ThreadedComment',
                     'sites.Site',
                     'redirects.Redirect',)

(I simply cut and pasted from the default `ADMIN_MENU_ORDER`.)
But this had no effect because of an `ImportError` in `mezzanine.urls` (line 19-31):

    # Remove unwanted models from the admin that are installed by default with
    # third-party apps.
    for model in settings.ADMIN_REMOVAL:
        try:
            model = tuple(model.rsplit(".", 1))
            exec("from %s import %s" % model)
        except ImportError:
            pass
        else:
            try:
                admin.site.unregister(eval(model[1]))
            except NotRegistered:
                pass

So I changed `ADMIN_REMOVAL` like this:

    ADMIN_REMOVAL = ('blog.models.BlogPost',
                     'generic.models.ThreadedComment',
                     'django.contrib.sites.models.Site',
                     'django.contrib.redirects.models.Redirect',)

But this still has no effect although there is no `ImportError` or `NotRegistered` in this code lines.

Then I did some debugging. The line 

                admin.site.unregister(eval(model[1]))

from `mezzanine/urls.py` (cited above) actually calls `mezzanine.boot.lazy_admin.unregister`, which looks like this:

    def unregister(self, *args, **kwargs):
        self._deferred.append(("unregister", args, kwargs))

It seems like after the boot process `mezzanine.boot.lazy_admin.lazy_registration` should be called, to apply all the deferred registering and unregistering:

    def lazy_registration(self):
        for name, deferred_args, deferred_kwargs in self._deferred:
[...]

PR #1867, opened 2018-08-08
Fixed ADMIN_REMOVAL setting

tiktuk:
Fixed `lazy_registration()` as pr. @skabbit's comment in #1313.


stephenmcd:
Thanks a lot. I think you moved (or didn't move) some comments in the code, and now they don't match the code they're commenting. Can you check that?

tiktuk:
Should be fixed now :) .

tiktuk:
I don't see why travis fails, it's not because of my fix, right?

stephenmcd:
> I don't see why travis fails, it's not because of my fix, right?

I think it's unrelated. 

BTW I missed your last comments so forgot about this, sorry about that.

Issue #1749, opened 2017-04-10
Django 2.0 support


solarissmoke:
Django 1.11 is out, and it would be great if we could update Mezzanine to support it. 

I'm happy to work on a PR but wanted to ask whether we can drop support for Django 1.8 at the same time (it makes life a lot easier if we don't have to maintain compatibility)?

ryneeverett:
Sorry but we want to continue supporting 1.8 since it's an LTS version. Our goal is to support all versions of django that are supported by upstream: https://www.djangoproject.com/download/#supported-versions. Looks like we *can* drop support for 1.9 though.

solarissmoke:
Okay, that's fair enough. I have got it working locally, hope to have a PR ready this weekend.

stanim:
Any news?

robd003:
Is anyone else having issues with context not being passed as a dict but rather as RequestContext?

stephenmcd:
@robd003 you'll find the merged PR for 1.11 support deals with that extensively: https://github.com/stephenmcd/mezzanine/pull/1750

jpic:
I think that is good to close until you want to release it first ?

keithdechant:
Is there going to be a new release including this change?

ushijimay:
I am looking forward to the new release, too.

you-zhou:
How could I start a Mezzanine with Cartridge project that uses Python 3.6 + Django 1.11LTS?
Is there a time line to for the new release? Or is it going to be hard to migrate from Mezzanine 4.2.3? Cheers.

Beomjun:
Any updates?

stephenmcd:
**Nothing is stopping you from using Mezzanine with the latest version of Django right now.** Simply specify the latest commit in this git repo as your dependency. This is a very basic task when working with Python, so if you don't know how, searching for "pip git" will yield a wealth of information on the topic.

With that said, please refrain from asking when there will be a new release or if there are any updates. The only update would be that the release has been made. When that happens, this issue will get closed off. If it's still open, that means there hasn't been a release yet. It will happen when I can find the time, which is something that I cannot predict. This is all relatively trivial given the fact you can use Mezzanine with the latest Django version right now.

you-zhou:
Thanks, Stephen.

alex-polosky:
I just wanted to say thank you so much for this code change. I spent a while trying to figure out what was going wrong after I updated to Django 1.11, and then I was able to stumble upon this.

[...]

PR #1876, opened 2018-09-04
Fix imports from deprecated `urlresolvers` module

simenheg:
Importing from the `django.core.urlresolvers` module was deprecated in favor of `django.urls` in Django 1.10, and the former was removed entirely in Django 2.0. Import from `django.urls` instead.

### DIFF
--- a/mezzanine/accounts/tests.py
+++ b/mezzanine/accounts/tests.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 from django.contrib.auth import get_user_model
 from django.contrib.auth.tokens import default_token_generator
 from django.core import mail
-from django.core.urlresolvers import reverse
+from django.urls import reverse
 from django.forms.fields import DateField, DateTimeField
 from django.utils.http import int_to_base36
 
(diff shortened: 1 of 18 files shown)


simenheg:
@stephenmcd, is it okay to drop the compatibility code for Django 1.9, like @jerivas suggested for filebrowser-safe in https://github.com/stephenmcd/filebrowser-safe/pull/114? If so I'll make an update to the commit.

stephenmcd:
@simenheg please go for it, thank you

simenheg:
> @simenheg please go for it, thank you

OK, I've updated it.

stephenmcd:
Thanks a lot

stephenmcd:
Hi @simenheg - it looks like you're doing a lot of work to get things working with 2.0, thanks for contributing back. Please let us know when you think everything works with 2.0 and we can potentially close off #1749.

simenheg:
> Please let us know when you think everything works with 2.0 and we can potentially close off #1749.

Yep, I will.

PR #1877, opened 2018-09-09
Fix failure in page processor test

simenheg:
Commit 4e94ab99 broke the `PagesTests.test_exact_page_processor_for` test, because `resolver_match` isn't always set on the `request` object. Check for that before attempting to set it to unbreak the test.


stephenmcd:
Thanks a lot

PR #1878, opened 2018-09-09
Make the search engine work in Django 2

simenheg:
The `*args` and `**kwargs` arguments to `QuerySet._clone` are removed in Django 2. Preserve the queryset attributes by copying them directly instead.

### DIFF
--- a/mezzanine/core/managers.py
+++ b/mezzanine/core/managers.py
@@ -171,13 +171,15 @@ class SearchableQuerySet(QuerySet):
             queryset = queryset.filter(reduce(ior, optional))
         return queryset.distinct()
 
-    def _clone(self, *args, **kwargs):
+    def _clone(self):
         """
         Ensure attributes are copied to subsequent queries.
         """
-        for attr in ("_search_terms", "_search_fields", "_search_ordered"):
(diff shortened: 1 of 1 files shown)


mezzanine/core/managers.py
```diff
@@ -171,13 +171,18 @@ def search(self, query, search_fields=None):
...
         Ensure attributes are copied to subsequent queries.
         """
-        for attr in ("_search_terms", "_search_fields", "_search_ordered"):
-            kwargs[attr] = getattr(self, attr)
-        return super(SearchableQuerySet, self)._clone(*args, **kwargs)
+        search_terms = self._search_terms
+        search_fields = self._search_fields
+        search_ordered = self._search_ordered
+        clone = super(SearchableQuerySet, self)._clone()
+        clone._search_terms = search_terms
```
stephenmcd: Can't these just be assigned directly with vars? eg `clone._search_terms = self._search_terms`

simenheg: Yep, they can. I've pushed an update.

stephenmcd: Excellent, thank you

PR #1879, opened 2018-09-09
Use class-based version of JavaScript catalog view

simenheg:
The function-based JavaScript catalog view was deprecated in favor of a class-based one in Django 1.10, and the former was removed entirely in Django 2.0. Use the class-based one instead.


stephenmcd:
Thanks

PR #1880, opened 2018-09-09
Fix invalid escape sequence in URL pattern

simenheg:
Invalid escape sequences are deprecated in Python 3.6. One such instance was found in the sitemap URL pattern; use a raw string literal there instead.


stephenmcd:
Thank you

PR #1881, opened 2018-09-18
Make test pass after September 2018

simenheg:
In September 2018, the search score of a page from 2016 with two matches surpassed that of a page from 2017 with one match, due to the non-linear age scale scoring, causing a test failure.

Update the test so that the page with one match is new enough to always score higher than the one from 2016.


stephenmcd:
Thank you!

PR #1882, opened 2018-09-19
Bump the required Django version range

simenheg:
The oldest still-supported Django release is currently 1.11, which will stay supported until early 2020. Drop Mezzanine's support for Django versions older than that.

Also include the Django 2.0 release series in the travis tests, since support for Django 2.0 is targeted for the next Mezzanine release.


stephenmcd:
Thank you

PR #1885, opened 2018-09-21
Fixed Deprecation Warnings for Django 2.0/2.1 and some Python deprecation warnings

vstoykov:
Two of the commits are for Django 2.0 and 2.1 compatibility. 

The firs one ("Regexes should be RAW all properly escaped") was needed to reduce the number of deprecation warnings which I see when run Python with `-Wdefault`.

If you think that this is too much noise I can separate it in another merge request.


vstoykov:
Tests are failing because of `filebrowser_safe`. There should be new version which supports Django 2.0 in order Mezzanine tests to pass. The problem is that failbrowser_safe CI depends on Mezzanine in order to pass. It's like chicken and egg problem.

vstoykov:
From the failing tests in #1886 I see that none of the changes here is required for Django 2.0/2.1. Probably after merging #1886 and #1887 and resolving the last failing test then I can rebase these changes to be properly tested especially regular expressions.

stephenmcd:
Thanks a lot

PR #1886, opened 2018-09-23
Feature: Testing through tox

vstoykov:
-  Allow local testing trough tox
- Change travis configuration to use tox for running the tests

My reasoning for this Pull Request is to allow other contributors to easily run tests locally for all supported configurations. I changed travis to use tox in order not to duplicate configuration for setup the test environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 .idea/
 local_settings.py
 .eggs/
+.tox/
 Mezzanine.egg-info/
 build/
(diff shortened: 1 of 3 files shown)


vstoykov:
Currently tests for Django 2.0 and Django 2.1 are failing because of two tests. I can make separate PR after this is merged to fix them. I added `filebrowser_safe` as dependency from git master in order for test suite to even run on Django 2+.

PR #1888, opened 2018-09-24
Fix: Django 2 (Python 3) Tests

vstoykov:
In #1886 was shown that only two tests are failing for Django 2+. The first is fixed in #1887. This is fix for the second one.

Because Django 2 is Python 3 only some of the compatibility shims in Django for converting byte strings to unicode strings was removed and code which uses the old behaviour will no longer works. Generally In Python 3 always native (unicode) strings will be used or bytes will be converted to strings with appropriate encoding.

For that reason I'm disabling `test_bytes_conversion` test which uses old Django's automatic conversion from bytes to unicode. When this test is disabled under Python 3 then the failing test `test_settings` under Django 2+ now pass correctly.

This is needed for #1749

### DIFF
--- a/mezzanine/conf/tests.py
+++ b/mezzanine/conf/tests.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 from future.builtins import bytes, str
 
+import sys
 from unittest import skipUnless
 import warnings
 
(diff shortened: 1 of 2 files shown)


stephenmcd:
Thank you

PR #1889, opened 2018-09-25
Remove usage of deprecated `allow_tags` attribute

simenheg:
The `allow_tags` attribute on methods of `ModelAdmin` was deprecated in Django 1.9, and removed in Django 2.0. Use `django.utils.html.format_html` instead to achieve the same effect.


stephenmcd:
Thank you

PR #1892, opened 2018-10-02
Update override_current_site_id to handle exceptions 

Souloist:
Currently, when `override_current_site_id` wraps a code block which raises an exception, it does not delete `override_current_site_id.thread_local.site_id`. 

Problem example: When working with celery, if the worker thread picks up a task which raises an exception, future tasks which call `current_site_id` will always return the first task's `site_id` since `override_current_site_id.thread_local.site_id` is still set.

Now context manager catches exception, cleans up after itself, and re-raises.


stephenmcd:
Nice, thank you

PR #1896, opened 2018-10-11
Utilise the latest TinyMCE library instead of older library

Olorin92:
This fork attempts to implement the newer TinyMCE library and make it work with the existing Mezzanine implementation for the older (vulnerable) TinyMCE.

Light testing shows that this should still work for the editor - but I'm not sure of any other more wide-reaching effects.

Comments/suggestions are welcome!


stephenmcd:
Thanks a lot for working on this. The functionality seems to work correctly in the admin as well as front-end editing. There are some styling issues in the admin, compare the two screenshots using the old and new versions:

Old:
screen shot 2018-10-16 at 2 54 51 pm

New:
screen shot 2018-10-16 at 2 55 03 pm

You can see the new editor blows out into the left-hand column where field names are listed.

Also the full-screen button fails badly both in the admin and front-end editing, in different ways for each.

Olorin92:
Great thanks for checking this - I figured someone else would now how to make it break!

I'll take a look at both of those things - just by eyeballing it it seems the newer TinyMCE has styling changes in that toolbar, so I suspect some of those changes are what's causing the weirdness.

Will commit changes once I fix those two things up.

Olorin92:
OK finally got the chance to start looking at this again - I can see what you're talking about with the full screen issue, but I'm not sure about the labels. For me the label just appears on top as normal:

image

Is this how it should look/different to what you're getting?

Fullscreen issue looks like the top of the editor sits underneath the top bar, so you can't undo the full screen without fiddling around. I'll see if there's an easy CSS change that prevents that from happening.

PR #1906, opened 2018-12-12
Upgrade to TinyMCE 4.9.1

gual:
When enabling [imagetools plugin](https://www.tiny.cloud/docs/plugins/imagetools/) for current version of TinyMCE (4.1.0) I got the error `module definition dependecy not found: tinymce/util/Promise`. Upgrading to TinyMCE 4.9.1 made the plugin and the editor work as expected. Tests results on `OK`.

PS: I followed the guidelines, let me know if there is something else required or something I missed. Thanks!


stephenmcd:
#1896 is open with a slightly older version - that PR contained some visual issues that needed to be resolved. Can you confirm these aren't present with this PR, and if they are, resolve them first?

Thanks a lot.

gual:
@stephenmcd I checked #1896.

## Fullscreen issue
I wasn't able to reproduce any issue on my branch. I get the same behavior with the current tinymce version as with the new one (the expected one).

Current: 
anadir_publicacion___mezzanine_and_upgrade_to_tinymce_4_9_1_by_gual_ _pull_request__1906_ _stephenmcd_mezzanine

Updated:
anadir_publicacion___mezzanine

## Styling issues
I did had the same issue of the editor blowing out into the left-hand column where field names were listed. The reason behind this change was that the new default value for width was `100%` and previously it had `auto`. I added the width option explicit on the default setup and voila...

anadir_publicacion___mezzanine

gual:
As for the

## Front-end Editing
I get the same behavior with the new version.

Current:
blog___funde

Updated:
blog___funde

PR #1916, opened 2019-04-14
Use Field.value_from_object() for Django 2.x

eht16:
The previously used private method "Field._get_val_from_obj()" was
removed in Django 2.0 and so the "dumpdata" management command
was broken. Django already offers the replacement method for a long
time.


eht16:
For reference: https://code.djangoproject.com/ticket/24716

jerivas:
Fixed in #1980

PR #1933, opened 2019-07-08
Compatibility with django >= 2.1

christianwgd:
- Django 2.1 removed the 'host' parameter from is_safe_url(), 'allowed_hosts' now is a required argument.
- As stated here https://github.com/django/django/pull/9938 in Django 2.1 the TOKEN_* constants were replaced by TokenType enums.


christianwgd:
As there ar two files in this pull request, which are not relevant to the stated changes, i close this and make a new pull request, sorry.

PR #1934, opened 2019-07-08
Compatibility with django >= 2.1

christianwgd:
- Django 2.1 removed the 'host' parameter from is_safe_url(), 'allowed_hosts' now is a required argument.
- As stated here https://github.com/django/django/pull/9938 in Django 2.1 the TOKEN_* constants were replaced by TokenType enums.


christianwgd:
One thing more: I'm afraid i did't respect python2 compatibility for the inspect.getfullargspec call. Will add that...

eht16:
Yay!
Though some of the commits rather look like spam and probably should be removed (just squash them away).

Together with https://github.com/stephenmcd/mezzanine/pull/1916 we should get a big step toward Django 2.2 support.

eht16:
Just noticed: https://github.com/fermorltd/mezzanine/commit/f6da0a2219fa4d1b7bce4cac4e47d864dab08844 is also necessary.

jerivas:
#1956 has been merged with support for Django >= 2.2

PR #1938, opened 2019-08-03
Fix __admin_media_prefix__ pointing to nonexistent resource

### LINKED ISSUE #1895: Error in admin/base_site.html when using ManifestStaticFilesStorage
`base_site.html` has this code: `    window.__admin_media_prefix__ = '{% static "admin" %}/';`, which tries to get the URL of a static file that doesn't exist ('admin'). This causes ManifestStaticFilesStorage to raise an error:

```
  File "django/contrib/staticfiles/storage.py", line 432, in stored_name
    raise ValueError("Missing staticfiles manifest entry for '%s'" % clean_name)
ValueError: Missing staticfiles manifest entry for 'admin'

```

bors-ltd:
Closes #1895

It breaks sites on production that use the ManifestStaticFilesStorage.

For such cases, Django provides the "get_static_prefix" template tag.

But it's an incitement to bypass the storage mechanism to produce the
URL to cache (by appending a hash in this case).

And I don't understand how it could work with more "exotic" yet valid
backends relying on cloud storage or CDNs, where the source file path
and the URL path won't match one to one.

I advise against this usage and call for its deprecation.

### DIFF
--- a/mezzanine/core/templates/admin/base_site.html
+++ b/mezzanine/core/templates/admin/base_site.html
@@ -1,5 +1,5 @@
 {% extends "admin/base.html" %}
-{% load mezzanine_tags i18n staticfiles %}
+{% load mezzanine_tags i18n staticfiles static %}
 
 {% block title %}{{ title }} | Mezzanine{% endblock %}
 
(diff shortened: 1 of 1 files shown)


rapto:
I have had this same problem. This fix it.

rapto:
Is `__admin_media_prefix__` used at all?

jerivas:
Huh, looks like this is already in master: https://github.com/stephenmcd/mezzanine/blob/872b0cd068f1f0086f4004fc5b08cd20945d645b/mezzanine/core/templates/admin/base_site.html#L24

PR #1946, opened 2019-10-08
All the urlpatterns updated according to regex and Django 2 updates

thewolfcommander:
I have updated all the urlpatterns according the DJango 2 updates as i have changed urls that contain regex from url() to re_path() and the urls that doesn't contain regex to path(). Also imported new modules for django.urls


jerivas:
Looks like this was handled by #1956

PR #1947, opened 2019-10-09
Fix issues with thumbnail template tag when using remote media storage

cjh79:
This is an implementation of the fix described here: https://groups.google.com/forum/#!msg/mezzanine-users/WGVaNhD5vRs/7Ktw7-UkosgJ

Rather than using `os` functions to create thumbnails on the filesystem, create the thumbnail in memory and push it to the media storage using `default_storage`. This fixes the thumbnail tag when using remote media storage.

I have tested this using remote storage (S3), and local storage. I would appreciate a second set of eyes on the fix.


cjh79:
Closing this so I can open the PR from a different branch. Sorry.

PR #1951, opened 2019-10-18
Make thumbnail template tag be compatible with custom storage engines

wonderbeyond:
Make thumbnail template tag be compatible with custom storage engines

* Doesn't create thumbnails in local fs while using custom storage
* Fix existence check logic with custom storage


wonderbeyond:
Oh, sorry! I found two similar pull requests: https://github.com/stephenmcd/mezzanine/pull/1949 and https://github.com/stephenmcd/mezzanine/pull/1922

PR #1955, opened 2019-12-08
radio field support for dynamic_inline.js

ChrisKuper:
When using radio buttons in dynamic inlines this raised an 'Unhandled
field in dynamic_inline.js:' alert. This was fixed by dublicating the
'checkbox' case in the function 'anyFieldDirty' in 'dynamic_inline.js'


jerivas:
Thanks!

PR #1958, opened 2020-03-05
Update nestedSortable

notthatnathan:
Updating jQuery UI to address [CVE-2010-5312](https://nvd.nist.gov/vuln/detail/CVE-2010-5312) breaks list sorting due to `$.ui.isOverAxis` being removed. This update to nestedSortable 2.1a fixes sorting.


notthatnathan:
Closing. Needs to be done, but needs the jQuery UI update, too, which I didn't include in this PR.

PR #1966, opened 2020-08-14
Update Mezzanine from Bootstrap 3 to Bootstrap 5 alpha 

hopechan:
Changes made so far

- Update bootstrap css and js files

- Remove jquery files

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,7 +61,7 @@ universal = 1
 
 [tox:tox]
 envlist =
-    py{36,37,38}-dj{22,30,31}
+    py{36,37,38}-dj{22,30,31,master}
     flake8
     black
     package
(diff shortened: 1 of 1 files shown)


mezzanine/core/static/mezzanine/js/jquery-ui-1.12.1.js
```diff
@@ -1,18706 +0,0 @@
-/*! jQuery UI - v1.12.1 - 2016-09-14
```
jerivas, line 1: jQuery is still being used in the admin and on the inline editor. As you add or remove files make sure you test the admin (specifically the admin page tree and admin inlines). You will also need to test the frontend editor by login in as admin and using the little "Edit" tags (see screenshot). This feature is documented here: http://mezzanine.jupo.org/docs/inline-editing.html

We definitely want to get rid of jQuery for regular users, but admin users will still need to load it for the inline-editor.

image

jerivas:
@hopechan thank you for contributing this! I see for now you are just removing the jQuery dependencies so I've left a review with some more detail on that

Issue #1977, opened 2021-01-26
dumpdata not working on latest commit (Django 2.2.1)


dbeltra:
When trying to dump the database with `python manage.py dumpdata` I get the following result:

`CommandError: Unable to serialize database: 'MenusField' object has no attribute '_get_val_from_obj'`

It looks like `_get_val_from_obj` has been deprecated since django 2.0, so it breaks on the latest mezzanine commit

jerivas:
Hello,
Can you share the complete traceback please? Thanks

dbeltra:
> Hello,
> Can you share the complete traceback please? Thanks

Sure, here it is :)

```
(venv) david at macbookpro in ~/Projects/my_website $ python manage.py dumpdata --traceback > data.json
Traceback (most recent call last):
  File "manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/management/commands/dumpdata.py", line 186, in handle
    object_count=object_count,
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/serializers/__init__.py", line 128, in serialize
    s.serialize(queryset, **options)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/serializers/base.py", line 107, in serialize
    self.handle_field(obj, field)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/serializers/python.py", line 50, in handle_field
    self._current[field.name] = self._value_from_field(obj, field)
  File "/Users/david/Projects/my_website/venv/lib/python3.7/site-packages/django/core/serializers/python.py", line 47, in _value_from_field
[...]

PR #1980, opened 2021-03-01
Replace _get_val_from_obj with value_from_object

petedermott:
Fix for #1977, I believe this functionality was just duplicated functionality and the method has been replaced.

See https://code.djangoproject.com/ticket/24716 for more info

### DIFF
--- a/mezzanine/core/fields.py
+++ b/mezzanine/core/fields.py
@@ -94,7 +94,7 @@ class MultiChoiceField(models.CharField):
             raise ValidationError(error)
 
     def value_to_string(self, obj):
-        value = self._get_val_from_obj(obj)
+        value = self.value_to_string(obj)
         return ",".join(value)
 
 


mezzanine/core/fields.py
```diff
@@ -94,7 +94,7 @@ def validate(self, value, instance):
             raise ValidationError(error)
 
     def value_to_string(self, obj):
-        value = self._get_val_from_obj(obj)
+        value = self.value_to_string(obj)
```
aparakian, line 97: Typo? Wasn't it supposed to be `value_from_object`?

ctrengove: It was suppose to be value_from_object. The change made now causes an infinite recursion when serializing the field (for example, with dumpdata).

aparakian: Made a PR for this #1999

Issue #1868, opened 2018-08-08
Enforce uniqueness of user email address via /admin/auth/user/x/ form


jogwen:
I have a Mezzanine project with ACCOUNTS_NO_USERNAME=True so that users sign up and log in using an email address and password, never their username. This was working great, and all front-end forms (sign-up and profile update) enforce uniqueness of the email address across the User table. 

However, two days ago, an user of the /admin/ screens edited an user's email address to match that of another user (long story, but it was a reasonable mistake to make as both accounts did in fact represent one real-world person). When the user in question next attempted to log in they got an error response and we got an error email showing, as one would expect, "MultipleObjectsReturned: get() returned more than one User -- it returned 2!".

Users of /admin/ shouldn't have to bear the burden of maintaining the integrity of the email address field when it is needed to uniquely identify an user object.

jerivas:
I agree. Would you like to set up a PR to contribute this change?

jogwen:
Many thanks @stephenmcd!

PR #1984, opened 2021-04-13
Don't enforce email uniqueness when blank, account for multiple existing dupes

dbischof:
Don't enforce email uniqueness when blank (emails can be optional), and account for possibility of multiple dupes (such as when migrating from prior to enforcement of uniqueness). Fix regression in stephenmcd/mezzanine#1868

### DIFF
--- a/mezzanine/core/admin.py
+++ b/mezzanine/core/admin.py
@@ -406,11 +406,12 @@ class SitePermissionInline(admin.TabularInline):
 class SitePermissionUserAdminForm(UserAdmin.form):
     def clean_email(form):
         email = form.cleaned_data.get("email")
-        try:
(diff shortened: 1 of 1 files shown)


jerivas:
Fixed in #2018

dbischof:
@jerivas, #2018 handles multiple existing dupes but it does not handle not enforcing email uniqueness when blank. 

[Django allows for blank emails](https://github.com/django/django/blob/7119f40c9881666b6f9b5cf7df09ee1d21cc8344/django/contrib/auth/models.py#L358). The code as it stands will not allow you to add/modify a new user with a blank email if one already exists (e.g. when using usernames and not emails for logins). 

Blank emails should be exempt from the uniqueness check.

jerivas:
Thanks for following up, I'll fix that next

Issue #1985, opened 2021-05-01
Pillow pinned to insecure version


gherkz:
The current pinning of Pillow `>= 7, <8` results in an insecure version of Pillow being used:
```
$ pipenv check
Checking PEP 508 requirements...
Passed!
Checking installed package safety...
40264: pillow <8.0.1 resolved (7.2.0 installed)!
Pillow 8.0.1 updates FreeType used in binary wheels to 2.10.4 to fix CVE-2020-15999.

40265: pillow <8.1.0 resolved (7.2.0 installed)!
Pillow 8.1.0 fixes TIFF OOB Write error. CVE-2020-35654 #5175.

40270: pillow <8.1.0 resolved (7.2.0 installed)!
Pillow 8.1.0 includes a fix for Read Overflow in PCX Decoding. CVE-2020-35653 #5174.

40271: pillow <8.1.0 resolved (7.2.0 installed)!
Pillow 8.1.0 includes a fix for SGI Decode buffer overrun. CVE-2020-35655 #5173.

40266: pillow <8.1.1 resolved (7.2.0 installed)!
Pillow 8.1.1 uses more specific regex chars to prevent ReDoS. CVE-2021-25292.

40272: pillow <8.1.1 resolved (7.2.0 installed)!
Pillow 8.1.1 fixes OOB Read in TiffDecode.c, and checks the tile validity before reading. CVE-2021-25291.

40273: pillow <8.1.1 resolved (7.2.0 installed)!
Pillow 8.1.1 fixes OOB read in SgiRleDecode.c. CVE-2021-25293.

40274: pillow <8.1.1 resolved (7.2.0 installed)!
Pillow 8.1.1 fixes incorrect error code checking in TiffDecode.c. CVE-2021-25289.

40275: pillow <8.1.1 resolved (7.2.0 installed)!
Pillow 8.1.1 fixes negative size read in TiffDecode.c. CVE-2021-25290.

40263: pillow <8.1.2 resolved (7.2.0 installed)!
Pillow 8.1.2 fixes memory DOS in BLP Image Plugin (CVE-2021-27921).

40267: pillow <8.1.2 resolved (7.2.0 installed)!
Pillow 8.1.2 fixes memory DOS in ICNS Image Plugin (CVE-2021-27922).

40268: pillow <8.1.2 resolved (7.2.0 installed)!
Pillow 8.1.2 fixes memory DOS in ICO Image Plugin (CVE-2021-27923).

[...]

PR #1986, opened 2021-05-01
Increase maximum version of Pillow

gherkz:
Addresses multiple CVEs (listed in https://github.com/stephenmcd/mezzanine/issues/1985).

I couldn't work out how to successfully run the tests, so can't confirm if this doesn't break anything.


jerivas:
Thanks! I've gone with less strict version requirements in general. See https://github.com/stephenmcd/mezzanine/commit/721847e84bf2af8bb31e2877bace827e8ae6655d

PR #1987, opened 2021-05-04
class*="collapse" matches even collapse-closed, unwanted behavior

dbischof:
In:
```
$('fieldset[class*="collapse"]').each(function() {
    $(this).addClass("collapse-open");
    $(this).removeClass("collapse");
});
```

`$('fieldset[class*="collapse-closed"]')` matches `<fieldset class="module aligned collapse-closed"/>`, which leads it to add the `collapse-open` class. This in turn means `collapse-closed` and `collapse-open` always toggle in tandem rather than flopping between each other as intended.


jerivas:
I'm unable to reproduce this on the latest master branch. Can you provide your admin class `fieldsets`?

PR #1988, opened 2021-05-05
fix: removed deprecated _get_val_from_obj

fepe55:
In 2.0 Django deprecated `_get_val_from_obj`[1] and should be replaced with `value_from_object`

[1]: See https://code.djangoproject.com/ticket/24716


jerivas:
Fixed in #1980

Issue #1989, opened 2021-06-01
Tests that modify settings and render pages fail on Django 3+


jerivas:
The test [`test_login_required`](https://github.com/stephenmcd/mezzanine/blob/master/tests/test_pages.py#L170) is currently failing at the point where settings are overridden:

```python
with override_settings(LOGIN_URL="login"):
    # Note: The "login" is a pattern name in accounts.urls.
    response = self.client.get(public_url, follow=True) # <------------------ Fails here
    self.assertEqual(response.status_code, 200)
    response = self.client.get(private_url, follow=True)
    self.assertRedirects(response, login)
```
```
mezzanine/template/__init__.py:53: in render
    context[parts[-1]] = tag_func(*args, **kwargs)
mezzanine/twitter/templatetags/twitter_tags.py:63: in tweets_default
    query_type = settings.TWITTER_DEFAULT_QUERY_TYPE
mezzanine/conf/__init__.py:259: in __getattr__
    editable_cache = self._get_editable(request=self._current_request)
mezzanine/conf/__init__.py:163: in _get_editable
    editable_settings = self._editable_caches[request]
mezzanine/conf/__init__.py:254: in __getattr__
    return getattr(django_settings, name)
.tox/py39-dj30/lib/python3.9/site-packages/django/conf/__init__.py:77: in __getattr__
    val = getattr(self._wrapped, name)
.tox/py39-dj30/lib/python3.9/site-packages/django/conf/__init__.py:204: in __getattr__
    raise AttributeError
E   AttributeError
```

From my investigation it looks like when `get` is called the `twitter_default` tag is called somewhere in the template, which in turns reads from the settings, and that crashes the site witn an `AttributeError` for `TWITTER_DEFAULT_QUERY_TYPE`. It seems to me `override_settings` is breaking Mezzanine's editable settings machinery (specifically the `_editable_caches` mentioned in the traceback).

PR #1994, opened 2021-08-17
Fix attribute error when calling clear_cache on settings.

lougeo:
Small fix to compliment a recent bug fix related to commit: 99a0441 issue:  #1989 

An AttributeError is thrown upon initial instantiation in places where settings.clear_cache() is called, due the removal of the init method.

This is my first ever PR to an open source project so let me know if I need to add/change anything.


jerivas:
Excellent contribution! Thanks

Issue #1964, opened 2020-08-07
When I logout, No Blog Post and Pages are showing on the home page


amannguptaggg:
I have changes the default page setting to blog only URL.
Now After post publish when I logout, all the pages I created and blog post are gone.

When I login again all the content are showing again.

Someone help what's the problem.

kenbolton:
Sounds like the "status" is set to draft and not published.

amannguptaggg:
no... status is published.
and for page unchecked for login required to see page

amannguptaggg:
I reinstalled Mezzanine from scratch and got error
`return Database.Cursor.execute(self, query)
django.db.utils.OperationalError: Problem installing fixtures: no such table: pages_page__old`

jerivas:
Hi @amannguptaggg, could you verify if the issue persist in mezzanine 5.0.0rc1? You can install it with `pip install mezzanine==5.0.0rc1`. If it does, please provide the exact steps to reproduce the issue on a blank Mezzanine installation. Thanks!

PR #1998, opened 2021-08-30
issue fixed '#1964'

ankur25112000:
do it before running server


jerivas:
Thanks for your contribution. Unfortunately I don't think this is the right approach as it would basically remove all content from the database (related to blog posts or not).

Feel free to add it as a workaround on the related issue.

PR #2012, opened 2021-12-30
Handle non integer values passed to `object_pk` and `replied_to`

Remiz:
When pentesters look for vulnerabilities they attempt to stuff random values to every forms (including the ratings and comments). This change prevents returning a 500 error when a non integer value is passed to `object_pk` in a rating or comment form.


jerivas:
Thanks!

PR #2031, opened 2022-04-08
fix: pin bleach to version less than 5

henri-hulski:
Bleach 5.0.0 doesn't except the keyword argument 'styles' for `clean()` anymore.
So we get the error `clean() got an unexpected keyword argument 'styles'`.
So I pinned the Bleach version to `bleach >= 2, <5`.
Bleach 5.0.0 also doesn't support Python 3.6 anymore. 
See https://github.com/mozilla/bleach/blob/main/CHANGES.


henri-hulski:
I had also to upgrade black to 22.3.0 and rerun it, as the old version was broken in the CI.
See psf/black#2964.

jerivas:
Could we switch to bleach > 5 instead? I would prefer to stay up to date instead of locking ourselves into an old version

henri-hulski:
Sure we can try.
Probably this will be a breaking change.
You could also merge this first to make a patch release to have a working version of Mezzanine 5 and I will prepare the move to bleach 5, which will be a major release - Mezzanine 6.

jerivas:
Well, I made a huge mess while trying to merge this from my phone (currently on vacation). I'll wait until next week to merge into `stable`.

Thanks for contributing!

henri-hulski:
What do you mean? It looks quite ok.
Can I use the master branch now for the next PR or do you want to change something?

PR #2032, opened 2022-04-12
Upgrade bleach to version 5 and drop Python 3.6 support

henri-hulski:
Bleach 5 doesn't directly support css sanitizing anymore, but it allows to use an external sanitizer.
The default sanitizer is tinycss2, which get installed with `pip install bleach[css]`.
I did also some clean-up in html.py.


henri-hulski:
@jerivas Bleach 5 doesn't support Python 3.6, so I had to drop it.
So it's actually a breaking change. The best would be as I said before to first release my last PR as a patch release and after this one as major release.

PR #2033, opened 2022-04-19
add a minified version of the collapse so non-debug mode works

### LINKED ISSUE #2005: [BUG] Collapsibles aren't working when DEBUG = False in Mezzanine 5.0.0
### Is there an existing issue for this?

- [X] I have searched the existing Issues

### Current Behavior

Collapsible sections of the admin aren't working in Mezzanine 5.0.0 when DEBUG = False. They do work as expected when DEBUG = True.

### Expected Behavior

I expect to be able to toggle a collapsible fieldset in the admin.

### Steps To Reproduce

1. Install Django at 2.2.25 and Mezzanine at 5.0.0 with Python 3.8.
2. Make sure the DEBUG setting = False.
3. Identify or create a page type that has a collapsible fieldset. I believe the Meta Data fieldset is collapsible by default, which is where I noticed it.
4. Load the page in the admin and observe that the fieldset cannot be toggled. It is grey by default and is not clickable. It should be blue and clickable to toggle open and closed.

### Environment

```markdown
- Operating System: Ubuntu
- Python version: 3.8
- Django version: 2.2.25
- Database engine and version: PostgreSQL@10
- Mezzanine version: 5.0.0
```

### Anything else?

For others with this issue, I resolved this temporarily for myself by reverting back to the 5.0.0a1 alpha release.

Before trying this, I also found that I could fix this by overriding the templates/admin/change_form.html with the following:

```html
{% extends "admin/change_form.html" %}

{% block extrahead %}
  {{ block.super }}
  {% comment %}
    After upgrading to Mezzanine 5.0.0, the collapse.min.js version of this file
    stopped working. I'm not entirely sure why, but pulling in the non-minified
    version seems to fix it.
  {% endcomment %}
  {% if not debug %}
    <script type="text/javascript" src="/static/admin/js/collapse.js"></script>
  {% endif %}
{% endblock %}
```

caffodian:
Fixes #2005 - Per https://github.com/stephenmcd/mezzanine/issues/2005#issuecomment-1061264104 , when Django is running in non-debug mode, it incorrectly tries to load Django's version of collapse.js, because we don't have a minified version. Since the file is pretty small I ended up just minifying it via the uglify website. I'm not sure if there's a preferred minifier here.

My prior notes:
---
this is how django decides which version of the collapse.js to load: https://github.com/django/django/blob/119f227aa62885f12cd7dd2558a62148d02adbb4/django/contrib/admin/options.py#L2327

Collapse.min.js is actually pre-minified and included in the Django package: https://github.com/django/django/blob/cab36618326524b336517c2c5fd62633916818e0/django/contrib/admin/static/admin/js/collapse.min.js

Since there's no minified version of that file included here, the minified Django collapse.min.js will be picked up instead of the unminified collapse.js here.

and i guess prior to this PR: https://github.com/stephenmcd/mezzanine/commit/95b138a961d5a102cfe79ed792cba1e9b93d134a it would have still worked because it'd load the minified django collapse + the unminified mezzanine collapse still.

---

Thanks!


caffodian:
@jerivas - I saw you looked at the issue for this (https://github.com/stephenmcd/mezzanine/issues/2005) - could you please check my explanation to see if it makes sense, and if there any chance this fix could be included in a release anytime soon?  (I'd like to avoid forking if possible, but we have some customers asking about this.)

thanks!

caffodian:
Oh wait!  This stopped being an issue in Django 3.2 because the collapse isn't in the list of js here: https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/options.py#L651

caffodian:
This might still be worth it for 2.2 compat reasons, but doesn't affect me personally anymore....sorry for the needless tag

henri-hulski:
Django 2.2 is already end-of-life.

PR #2035, opened 2022-04-29
Add and document collectstatic & collecttemplates.

poikilos:


henri-hulski:
Can you remove the linting changes as they don't belong here and are conflicting with `master` now.

Issue #2034, opened 2022-04-23
[BUG] "Fixed" URLs are still broken, which can harm your Google ranking.


Poikilos:
### Is there an existing issue for this?

- [X] I have searched the existing Issues
  - This is related to, but not the same as, issue #2022
  - This is related to, but not the same as, the issue of broken links on <https://pypi.org/project/Mezzanine/> which I reported over e-mail April 10, 2022 since there was no link to this repo there.

### Current Behavior

- The readme was changed to use the github page to replace the domain, but sub-urls such as are still broken.
  - When you click a link in the readme, a GitHub 404 page appears.
  - It looks bad for the project.
  - If the links are erased from the readme, the readme is less useful. Where to find the information from the broken links is unclear. 
- When you click any of several Google results (such as <https://github.com/stephenmcd/mezzanine/docs/packages.html>) about how to use mezzanine, a GitHub 404 page appears.
  - Google tracks this type of thing, so it will harm the ranking of your domain.

### Expected Behavior

- Get the helpful information from the search result of readme somehow.
- Make Google links work to prevent the domain from being downranked.

I didn't know where the official repo was so I mentioned over e-mail that <https://pypi.org/project/Mezzanine/> had bad links. That was fixed since then. However, the following from my e-mail still remains true regarding the "Current Behavior" above:

> . . . https://jupo.org/ [says] ERR_CERT_COMMON_NAME_INVALID.
>
> Do you need hosting? I can probably host this for free on one of my VPSs if you think the site won't be overloaded with traffic, and throw in a domain. I can also help with configuring the domain properly such as using a free cert from https://letsencrypt.org/
> 
> I hope this helps.
> 
> I would like the site back not just so I can use it but also so I can use a tool that has a community of support rather than appearing abandoned. I'm trying to fix the old ninja-ide website and want to make it work with the latest mezzanine. If people don't feel that mezzanine is supported, fewer people will choose it, and therefore fewer people will report issues or submit pull requests. Therefore, even if you provide a URL to me, the site not existing is still a problem.

Plus, I actually want to be able to learn how to do the things to which the readme (used to) link, not just have the links deleted.

### Steps To Reproduce

1. Click any link in the readme or via a search engine that have a "/" after the main site.
[...]

Issue #2030, opened 2022-04-07
Point documentation links to Read the Docs


jerivas:
### Discussed in https://github.com/stephenmcd/mezzanine/discussions/2029

<div type='discussions-op-text'>

<sup>Originally posted by **blitzd** March 28, 2022</sup>
Hey - the documentation site seems to be offline since sometime mid last week. Just a heads up - http://mezzanine.jupo.org/sites/
</div>

Links should now point to https://mezzanine.readthedocs.io/en/latest/

poikilos:
As I suggested at #2034, pointing to the section of the correct section of the docs (instead of just the root of the docs, and instead of search-and-replace for the domain) would ensure that the readme remains as useful as it was when the links went to the correct instructions.

denisemauldin:
@jerivas The https://mezzanine.readthedocs.io/en/latest/ shows the 4.3.1 documentation.  Can you update to latest please?

henri-hulski:
Actually I just added an automatic documentation generation to gh actions.
The documentation is in the gh-pages branch and to make it accessible under http://mezzanine.jupo.org/, we need only to activate gh-pages in the settings and create a CNAME record for mezzanine.jupo.org which points to stephenmcd.github.io.
The same for cartridge.
This has to be done by @stephenmcd, and I already mailed him still waiting for the answer.

henri-hulski:
Here the relevant parts of the mail, just to have the information here:

Actually I'm now adding the CNAME files directly to the gh-page branches through the gh action, so it should be enough if you create CNAME entries for mezzanine.jupo.org and cartridge.jupo.org which both point to stephenmcd.github.io.
Then activate gh-pages in the settings. You don't need to specify the custom domain as this is already done adding the CNAME file to the gh-pages branch.
After the certificate is ready you should also set "Enforce HTTPS".

henri-hulski:
Here are the docs for managing a custom domain for gh-pages: 
https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site

And here some details about the CNAME file which will be added to the gh-pages branch:
https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/troubleshooting-custom-domains-and-github-pages#cname-errors

As I recreate the branch within gh actions on each build the CNAME file will be re-added through the action.

bastianhilton:
[...]

PR #2041, opened 2022-07-02
Change documentation urls

ccurvey:
I think I got this right.  This should fix the problems with the initial site homepage containing dead links.  I think this will fix both #2030 and #2034 (if I understand the issues correctly)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -280,3 +280,4 @@
 * Chris Hawes
 * Andrii Soldatenko
 * Christian Kuper
+* Chris Curvey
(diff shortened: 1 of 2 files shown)


henri-hulski:
At the moment we automatically create the docs in the `gh-pages` branch.
So the solution would be to activate gh-pages in the settings and point `mezzanine.jupo.org` to it (see #2030).

Problem is that @stephenmcd is the person who has the permission and he is not responsive.

Alternatively we could set up an automatic update of readthedocs, but I don't know who has access here.

ccurvey:
Until we can get the CNAME records, or the appropriate permissions for readthedocs, would it make sense to just point those links back to the .RST files in GitHub?

damascene:
@ccurvey I just did a PR for that: https://github.com/stephenmcd/mezzanine/pull/2047

Issue #2054, opened 2023-02-28
[BUG] Bleach 6.0.0 breaks escape method at  mezzanine/utils/html.py


dbeltra:
### Is there an existing issue for this?

- [X] I have searched the existing Issues

### Current Behavior

When trying to save a rich text page you get the following error:

`TypeError: unsupported operand type(s) for +: 'frozenset' and 'list'`

### Expected Behavior

The page should be saved without errors

### Steps To Reproduce

1. Install Mezzanine 6.0.0 and bleach 6.0.0
2. Create a new project and try to save a page

### Environment

```markdown
* Mezzanine 6.0.0
* Django 4.1.7
* Python 3.8.13
* SQLite 3.38.5
* Darwin 21.4.0
```

### Anything else?

On the latest bleach release, ALLOWED_PROTOCOLS has been changed from list to a frozenset, this is the specific commit: https://github.com/mozilla/bleach/commit/29231a10eb983e25b59c8edc5b6abcb12dbaaabe

This makes this code (https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/utils/html.py#L113) crash:
`protocols=ALLOWED_PROTOCOLS + ["tel"]` since lists and frozensets can't be added

Downgrading to bleach==5.0.1 fixes the issue.

fioref:
same issue with:
* Mezzanine 6.0.0
* Django 4.1.7
* Python 3.10.9



> ### Is there an existing issue for this?
> 
>     * [x]  I have searched the existing Issues
> 
> 
> ### Current Behavior
> 
> When trying to save a rich text page you get the following error:
> 
> `TypeError: unsupported operand type(s) for +: 'frozenset' and 'list'`
> ### Expected Behavior
> 
> The page should be saved without errors
> ### Steps To Reproduce
> 
>     1. Install Mezzanine 6.0.0 and bleach 6.0.0
> 
>     2. Create a new project and try to save a page
> 
> 
> ### Environment
> 
> ```gfm
> * Mezzanine 6.0.0
> * Django 4.1.7
> * Python 3.8.13
> * SQLite 3.38.5
> * Darwin 21.4.0
> ```
> 
> ### Anything else?
> 
> On the latest bleach release, ALLOWED_PROTOCOLS has been changed from list to a frozenset, this is the specific commit: [mozilla/bleach@29231a1](https://github.com/mozilla/bleach/commit/29231a10eb983e25b59c8edc5b6abcb12dbaaabe)
> 
[...]

PR #2059, opened 2023-04-09
fix frozenset + list and unsupport styles parameter

DatCGI2net:
I found this bug  when trying to delete a sub-items on my homepage. 

- Exception Value: unsupported operand type(s) for +: 'frozenset' and 'list'


molokov:
See #2054 - but at least you have supplied a fix. Let's hope it gets merged.

henri-hulski (approved):

PR #2067, opened 2023-12-24
Accounts: For signup, login and password reset, username and password should be treated case insensitive

### LINKED ISSUE #2066: [BUG] accounts - username/email is not treated case insensitive
### Is there an existing issue for this?

- [X] I have searched the existing Issues

### Current Behavior

For accounts: usernames email addresses are matched case sensitive, which means the following:

a) Two accounts can be signed up for using the same email address but different case.

NOTE: Usernames are checked as case insensitive, so you can't create two accounts with the same username but different case ( https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/accounts/forms.py#L151-L156 )

b) When logging in using a username/email which doesn't match the case of an account, an error of "Invalid username/email and password" is given.
c) When attempting to reset a password via username/email which doesn't match the case of account, an error of "Invalid username/email" is given.

This has tripped up many users of my website, because they've not used all lowercase on their email addresses when signing up, and had forgotten their usernames for password reset or login. So of course, they attempted with email address and discovered that they couldn't get in that way either (because the case of the email wasn't an exact match).

Given that email addresses are cases insensitive, so should our use/authentication of them be here. We also disallow two usernames of different cases when creating an account, but username for login/password reset is case sensitive, which isn't consistent.

### Expected Behavior

Username and Email matching should be case insensitive for user accounts.

### Steps To Reproduce

A) Sign up for an two accounts with the same email address, just different case. This succeeds and two accounts are created. Expected error: "This email is already registered".

B) Log in to an account using an username or email with different case to the one registered. This should succeed.

C) Reset password on an account using an username or email with different case to the one registered. This should succeed.

### Environment

```markdown
- Operating System: Ubuntu 22.04
- Python version: 3.10.12
- Django version: 4.2.8
- Database engine and version: SQLite 3.37.2
- Mezzanine version: 6.0.0 (master)
```

### Anything else?

Should be a fairly simple fix - instead of 

```
username_or_email = Q(username=username) | Q(email=username)
```
we should use
```
username_or_email = Q(username__iexact=username) | Q(email__iexact=username)
```
This line appears at https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/accounts/forms.py#L264
and https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/auth_backends.py#L28

For signup, the line to fix would be: https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/accounts/forms.py#L184 for email addresses.

Note that *usernames* are checked case insensitive when signing up at https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/accounts/forms.py#L151-L156 so this should likely apply to usernames as well.

I'll make a fix on my fork and see if I can raise a pull request.

molokov:
Fixes #2066 

My apologies that this branch contains other unrelated fixes, but some of these are required to run with the latest versions of Bleach/PIL and thus are part of my fork already.


henri-hulski (approved):
Looks good to me. Is it ready to be merged?

molokov:
Unfortunately the PR has a few commits (from my master) related to other tickets, but if you are able to cherry pick the most recent two to merge, that should be enough for this one!

[4015354](https://github.com/stephenmcd/mezzanine/pull/2067/commits/40153547401dff610c4ee45a580902ff21c52d6e)
and 
[f10dc82](https://github.com/stephenmcd/mezzanine/pull/2067/commits/f10dc8233445a11237a0393b1151f02017e5249d)

mezzanine/core/templatetags/mezzanine_tags.py
```diff
@@ -429,12 +429,17 @@ def thumbnail(
             pad_container = Image.new("RGBA", pad_size, padding_color)
             pad_container.paste(image, (pad_left, pad_top))
             image = pad_container
+            # Make thumbnail a PNG - required if original isn't one
```
molokov, line 432: I think this change probably doesn't need to go back to main, I'm sure there's a better solution to changing a JPG's thumbnail to PNG (to allow for transparency background when using padding). Feel free to skip these lines on merge

henri-hulski (commented):
I think this change is breaking in existing projects where usernames with different cases are treated as 2 different accounts. Also in Django username and email are treated case sensitive. So this would be a quite substantial change. I'm not sure if at this point (mezzanine is rather in maintenance mode) we should make such a change.
I think to have them insensitive makes sense but if someone cares it should be solved on project base.
Instead of changing the behavior of mezzanine and make it different from Django we could rather document the problem.

henri-hulski:
So the "Changing ANTIALIAS to LANCZOS in thumbnail()" commit is still necessary if we want to allow for Pillow >= 10 and Python 12. Maybe I can cherry pick it.

molokov:
> I think this change is breaking in existing projects where usernames with different cases are treated as 2 different accounts. Also in Django username and email are treated case sensitive. So this would be a quite substantial change. I'm not sure if at this point (mezzanine is rather in maintenance mode) we should make such a change. I think to have them insensitive makes sense but if someone cares it should be solved on project base. Instead of changing the behavior of mezzanine and make it different from Django we could rather document the problem.


Fair enough - I do think case sensitivity in an email address is just plain wrong, I could accept it for usernames. Consistency with Django makes sense, however. 

No issues if you don't want to pull this back to master.

henri-hulski:
I agree with email addresses. I actually normalize them and store them lowercase in the database. So strictly spoken the RFC 5321 allow mail servers to treat the local part of the email address case sensitive.
I've merged the ANTIALIAS - > LANCZOS commit into master. Thank for your effort.

PR #2073, opened 2024-06-12
raises error on nested override site

rapto:
When calls to override_current_site_id are nested, no errors are raised until the deepest context `__exit__`. This patch detects this use and raises an error.

Of course, we could _implement_ nested calls. The test is prepared for that time.


henri-hulski (approved):
Thanks! Looks good to me.

PR #2084, opened 2026-04-17
Updates for Django 5.2 support

molokov:
Adding support/workflows for Django 5.0-5.2, with Python 3.10-3.14


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{38,39,310,311,312}-dj{22,30,31,32,40,41}
+    py{38,39,310,311,312,313,314}-dj{22,30,31,32,40,41,50,51,52}
```
henri-hulski: I think dj42 was missing here. Can you add it?

henri-hulski (approved):
Thanks for your work. Good that the project is still moving on!

henri-hulski:
I added a comment to `tox.ini`.  Can you add the missing `dj42`?

